PR #1, opened 2024-09-01
nix: initial cleanup; build & deploy docs

NotAShelf:
Cleans up most helper functions to avoid ambiguity in function argument order & argument names. The documentation derivation still needs more cleaning (lib is a *terrible* name to give to a helper function) but I ran out of time for today.

Also builds the docs with nix, and deploys via GH actions. Should be pretty straightforward, but untested. Website builds fine with Nix, and the result dir can be server with Python's http - everything is in order.

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -1,33 +1,38 @@
 {
-  pkgs,
-  astal,
+  flakePkgs,
+  stdenvNoCC,
+  formats,
+  gi-docgen,
+  glib,
+  json-glib,
+  gobject-introspection,
+  gtk3,
+  gtk-layer-shell,
+  gdk-pixbuf,
+  libdbusmenu-gtk3,
+  wireplumber,
+  networkmanager,
+  ...
 }: let
-  toTOML = (pkgs.formats.toml {}).generate;
+  toTOML = (formats.toml {}).generate;
 
-  gen = pkg: name: out: toml: ''
-    mkdir -p $out/${out}
-    gi-docgen generate -C ${toTOML name toml} ${astal.${pkg}.dev}/share/gir-1.0/${name}-0.1.gir
-    cp -r ${name}-0.1/* $out/${out}
+  genRefForPkg = {
+    name,
+    pkg,
+    outPath,
+    metaData,
+  }: let
+    tomlMetaData = toTOML name metaData;
+    devOutput = flakePkgs.${pkg}.dev;
+  in ''
+    mkdir -p $out/${outPath}
+
+    gi-docgen generate \
+      -C ${tomlMetaData} ${devOutput}/share/gir-1.0/${name}-0.1.gir
+
+    # Copy files
+    cp -rv ${name}-0.1/* $out/${outPath}
   '';
-
-  lib = name: namespace: description: {
-    authors ? "Aylur",
-    dependencies ? {},
-    out ? "libastal/${name}",
-  }:
-    gen name "Astal${namespace}" out {
-      library = {
-        inherit description authors;
-        license = "LGPL-2.1";
-        browse_url = "https://github.com/Aylur/Astal";
-        repository_url = "https://github.com/Aylur/Aylur.git";
-        website_url = "https://aylur.github.io/astal";
-      };
-      dependencies = {
-        inherit (dependency) "GObject-2.0";
-        inherit dependencies;
-      };
-    };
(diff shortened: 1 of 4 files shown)


docs/default.nix
```diff
@@ -1,33 +1,38 @@
...
-        repository_url = "https://github.com/Aylur/Aylur.git";
-        website_url = "https://aylur.github.io/astal";
-      };
-      dependencies = {
-        inherit (dependency) "GObject-2.0";
-        inherit dependencies;
-      };
-    };
+    gi-docgen generate \
+      -C ${tomlMetaData} ${devOutput}/share/gir-1.0/${name}-0.1.gir
```
fufexan, line 31: I assume the version should also be parameterized (maybe in the future).

docs/default.nix
```diff
@@ -36,9 +41,40 @@
       docs_url = "https://developer.gnome.org/gobject/stable";
     };
   };
+
+  lib = name: namespace: description: {
```
fufexan, line 45: Why not name this `docBuilder`?

flake.nix
```diff
@@ -5,48 +5,97 @@
     self,
     nixpkgs,
   }: let
-    version = builtins.replaceStrings ["\n"] [""] (builtins.readFile ./version);
+    astalLib = pkgs.callPackage ./nix/lib.nix {inherit version;};
+    inherit (astalLib) getVersionFromFile mkAstalPkg;
+
+    version = getVersionFromFile ./version;
     system = "x86_64-linux";
```
fufexan, line 12: Any reason this is x86_64-linux only?

docs/default.nix
Aylur: The gir api version is not going to change in the near future I suppose, but the astal lib api version might go to 1.0 at some point, but definitely not in the near future, its fine for now I think

flake.nix
Aylur: Thats the only machine I have, but if all the dependencies can be built on other architectures, they can be added too, I just did not bother to check.

docs/default.nix
NotAShelf: From my initial message:

> The documentation derivation still needs more cleaning (lib is a terrible name to give to a helper function) but I ran out of time for today.

I intend to redo that one as well (mainly for arguments, because I hate foo: bar: baz: for multiple arguments) but that will have to be for another time.

Aylur:
@NotAShelf If I understand correctly the gh action will publish both the astro site, and the generated gi-docgen site to `aylur.github.io/astal`?
I want the gi-docgen site to be availablte at `aylur.github.io/astal/reference` do I just set [target-folder](https://github.com/JamesIves/github-pages-deploy-action?tab=readme-ov-file#optional-choices) to `reference`?

NotAShelf:
This builds and serves the result of `nix build .#docs` as the root.

The astro build is not included, but it could be if we tell nix to output the reference docs (which `.#docs` is) in a directory nested in the astro build.

Aylur:
@NotAShelf Thanks for the help, but I am going to reject this PR.
I did not find a way to deploy and correctly serve the generated site from `gi-docgen`, but I have found a workaround by pushing directly to the pages repo.
For the nix changes, I know my nix code does not follow best practices, but tbh I don't care because its not a public API. For example I don't want to outsource functions to nix/lib as I feel it just adds noise to the repo and I also prefer positional arguments for `mkAstalPkg` because I want the list of libraries to be easy to read.
But I did make some changes you suggested.

I have one question though which I could not find an answer to online why the usage of [builtins.path](https://github.com/Aylur/astal/blob/ebfe47a0be6938f517f121a2b2dbc8f97967d6a8/docs/default.nix#L72C1-L75C7)?
What is the difference between `src = ./.` and `src = builtins.path { name = "src"; path = ./.; }`?

NotAShelf:
The usage of `builtins.path` is a little trick for making source paths reproducible while working with local paths. Helps avoid unnecessary rebuilds when a source path changes, but its contents does not. See the [nix.dev article](https://nix.dev/guides/best-practices#reproducible-source-paths) on this.

On another note, I would like to nudge you *gently* to use structured args (`{foo, bar}` instead of `foo: bar:`). I understand your reasoning, and I cannot (nor will) argue with it but there is only to gain from having a structured and readable codebase.

PR #9, opened 2024-09-14
Added examples for python and lua to the docs :nerd_face: 

tokyob0t:
uwu


Aylur:
thanks, I removed version numbers for a few reasons
* in the distance future when there would be an api version I would 100% forget about these
* I feel like they are redundant, because my end goal would be to have a dedicated section for each language going through things like these, so these examples' purpose is to just show what the typelib name is
* makes it look better to switch between views as its _almost_ line by line the exact same thing

PR #13, opened 2024-09-18
feat: add functions to accesspoint.vala

AhmedSaadi0:
Try adding to `accesspoint.vala`
- `connect_to_ap(string? password = null)`
- `forget_ap(string uuid)`

```json
{
  "saved": true,
  "uuid": "682fb234-4373-41cb-9dc6-df813a30hwl0",
  "hidden": false,
  "requiresPassword": true
}
```


AhmedSaadi0:
- Add `disconnect_from_ap()`

Aylur:
Thanks for the PR!
I still really don't understand much about network manager, whats a uuid used for and why does the `forget_ap` method requires it as a parameter?

AhmedSaadi0:
The UUID (Universally Unique Identifier) is used to uniquely identify a connection profile in NetworkManager's system, differentiating it from all other network profiles, regardless of network type (Wi-Fi, Ethernet, VPN, etc.).

The `forget_ap` method requires the UUID to ensure it correctly targets and removes the specific Wi-Fi network profile, even if there are multiple networks with the same SSID or other similar properties.

Aylur:
But each access point has only one UUID as I understand, so why does the forget_ap method requires it as a paramater? shouldnt it use the instance's uuid?

AhmedSaadi0:
You're right, the current implementation only supports one connection profile per access point. However, each profile corresponds to a unique network configuration, and `NetworkManager` allows you to create and store multiple profiles for the same Wi-Fi access point (or network). Which means, we need to create a `connection_profile.vala` file to manage the profiles associated with each access point, and move `forget_ap()` into it.

Right now I just used the instance's uuid until I get some extra time to create `connection_profile.vala`.

AhmedSaadi0:
### accesspoint.vala
- `public List<weak ConnectionProfile> profiles`
- `public ConnectionProfile? active_profile`
- `public async void deactivate_connection()`
```vala
public async void create_new_connection(
        string? password = null,
        string? ssid = this.ssid,
        string? bssid = this.bssid,
        string mode = "infrastructure",
        string? band = null,
        uint channel = 0,
        string? mac_address = null,
        bool powersave = true,
        string security_type = "wpa-psk",
        string auth_alg = "open",
        string ip4_method = "auto",
        string[]? ip4_dns = null,
        string ip6_method = "auto",
        bool autoconnect = false,
        bool read_only = false
    )
```
### accesspoint_profile.vala
- password
- uuid
- is_active
- `activate_connection()`
- `deactivate_connection()`
- `delete_profile()`

### TODO: `update_profile(args...)`

AhmedSaadi0:
I think everything is ready now

### accesspoint.vala
- hidden
- requires_password
- active_profile
- `create_full_connection(...full_profile_args)`
- `create_new_connection(string? password = null)`
- `deactivate_connection()`

### accesspoint_profile.vala
- uuid
- is_active
- id
- autoconnect_priority
- autoconnect
- visible
- ip4_method
- ip6_method
- mode
- channel
- mac_address
- powersave
- `get_psk()`
- `get_key()`
- `auth_algo()`
- `get_dns()`
- `activate_connection()`
- `deactivate_connection()`
- `delete_profile()`
- `update_profile(...full_args)`

gitmeED331:
Will be awesome to have these!

Jas-SinghFSU:
Do we know what's pending on this?

davfsa:
@Aylur really sorry for bothering, but would it be possible to get an update on this?

The current network implementation is quite unusable as it stands and very error prone (at least I have personally found)

Yilmaz4:
+1, I use Hyprpanel and its network module is broken now with NetworkManager version 1.50+, the developer said he's waiting for this PR to be merged

Also, does this expose WireGuard connections? #79 talks about this

Aylur:
Finally sat down and went through this and ended up hijacking this PR.

Some feedback:

- you should not use `Object.set_property`, just write the property

- `AstalNetwork.ConnectionProfile` is quite pointless: it does not add any functionality it just reads/writes properties

- adding a new connection each time is not desirable imo

Also just leaving this comment: The goal with this network library is to abstract as much as possible from network manager and expose a limited subset of its functionality which are the most used for desktop shells (think quick settings like list menu). If anyone plans to write a network settings app don't plan on using this lib, just use NM.

Example usage of this API in JS

```ts
const network = Network.get_default();

network.wifi.connect("state-changed", (_, _old, _new, reason) => {
  if (reason === NM.DeviceStateReason.NO_SECRETS) {
    console.log("wrong password");
  }
});

async function activate(ap: Network.AccessPoint) {
  let pw: string | null = null;

  if (ap.get_connections().length == 0 && ap.requiresPassword) {
    // popup asking for password
  }

  // in case its not "wpa-psk" you can use
  // `network.client.add_connection_async` bofore `ap.activate(null)`

  // or just simply 
  // ap.activate(pw, null) if you don't care about when it returns
  return new Promise((resolve, reject) => {
    ap.activate(pw, (_, res) => {
      try {
        resolve(ap.activate_finish(res));
      } catch (error) {
        reject(error);
      }
    });
  });
}
```

PR #16, opened 2024-09-22
docs: added some notes for /usr/local installation.

f3rmata:
Actually I'm not sure if it is a meson pkg-config problem.
But anyway, there's no harm to add some information on wiki.

### DIFF
--- a/docs/ags/installation.md
+++ b/docs/ags/installation.md
@@ -46,10 +46,18 @@ go install
 ```
 
 :::info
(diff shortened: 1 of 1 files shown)


Aylur:
The wording is technically not right, because this step is only required on some distros. For example Iirc Fedora for example has /usr/local already setup
Also, I would rather change the docs to not mention this at all and just recommend installing with `--prefix /usr` either way. I don't know if this applies to every distro, but most if not all of them uses /usr

PR #18, opened 2024-09-24
add idle inhibitor

kotontrion:
this Pr adds an inhibit property to Astal.Window. when this property is true, it will prevent the compositor from triggering the idle behavior, like suspend, or lock screen, as long as the window is visible.

added runtime dependencies:
wayland-client (already implicitly required as a dependency of GTK)
added build dependencies:
wayland-protocols
wayland-scanner

Todo:
~when the window hides and shows again, a new wl_surface is created for it, therefore a new inhibitor has to be created as well.~


Aylur:
Can we move the vapi dir into the src dir? Otherwise lgtm

kotontrion:
> Can we move the vapi dir into the src dir?

done

PR #22, opened 2024-09-26
Added keybind method to widget

tokyob0t:
<span style="display: inline;">
    Sorry if I messed up the indentation
    
</span>

### DIFF
--- a/core/lua/astal/widget.lua
+++ b/core/lua/astal/widget.lua
@@ -1,288 +1,334 @@
 local lgi = require("lgi")
 local Astal = lgi.require("Astal", "0.1")
 local Gtk = lgi.require("Gtk", "3.0")
+local Gdk = lgi.require("Gdk", "3.0")
 local GObject = lgi.require("GObject", "2.0")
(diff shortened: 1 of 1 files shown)


Aylur:
Firstly, this should be done in Vala, like the css and class name setters/getters

Secondly, I don't really see the point in this anymore. My reasoning for this in ags was that the user should not be required to import the `Gdk` namespace and connecting to signals had to be done in a `setup` callback, so the I tried to simplify it with the keybind method.
Now that importing namespaces have to be done a lot more (for example in the case of Gtk.Align and Astal.WindowAnchor), the first point no longer applies. Also, since prefixing a property with `on_` works both in Lua and JS as a signal handler the second point no longer applies either.

```lua
Button {
  on_key_press_event = function(self, event) {
    if event.mod == somemod1 + somemod2 and event.key == somekey then
    end
  }
}
```

```js
Button({
  onKeyPressEvent: (self, { key, mod }) {
    if (key == somekey && mod == somemod1 | somemod2) {
    }
  }
})
```

PR #25, opened 2024-09-28
core: fix astalified container not removing children

kotontrion:
If using astalify on a widget inheriting from container and binding the children, will result in new children being added by the bind, but the old ones are not removed. This PR fixes that.

### DIFF
--- a/core/gjs/src/astalify.ts
+++ b/core/gjs/src/astalify.ts
@@ -19,7 +19,9 @@ function setChildren(parent: Gtk.Widget, children: Gtk.Widget[]) {
         if (ch)
             parent.remove(ch)
     }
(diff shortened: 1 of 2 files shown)


Aylur:
This elseif clause should only run if `!(parent instance of Astal.Box || parent instanceof Astal.Stack)`

And I just realized we could outsource their `no_implicity_destroy` prop similar to the class_name and css props, so that everywidget cleans up after themselves. This would remove the need for 
image

For now just add the Box and Stack check and also add this to the [lua lib](https://github.com/Aylur/astal/blob/d9cdac25f78f4c9aaf0b12461a0d737dc392e746/core/lua/astal/widget.lua#L46) so we can merge it as a hotfix. I'll refactor it out later

PR #31, opened 2024-09-29
new lib: cava

kotontrion:
This PR adds a new library for audio visualization using cava.

It depends on [libcava](https://github.com/LukashonakV/cava), a fork of cava, which builds it as a shared library.

Todos:

- [x]  make properties changeable during runtime
- [x] add functionality to stop/start audio capture
- [x] docs (only integration in github action is missing)


Aylur:
Specifying a config-path in a library does not sound like a good solution.
Does `libcava` work this way by default? If so we should have a property on the cava object for each option and at runtime write it to a file `RUNTIME/astal/cava` and specify that for libcava under the hood.

kotontrion:
Cava provides a library cavacore, which you have to statically link against. But this only provides access to the audio processing engine. You have to get the audio samples yourself.

The cava fork i linked builds cava as a shared library and therefore can process audio samples from all kinds of different audio sources. Using a config file is not really necessary, as you can just set the struct members yourself, but it handles quite a few things like automatic source selection, so using a config file is an easy point to start. It could be removed later when other properties are implemented.

Or switch to cavacore, which is simpler to configure, but the audio input has to be implemented, which i could do, but maybe only support pipewire for now.

kotontrion:
I think it's done now. Except for the nix stuff.

The cava lib does terminate the program when some options have invalid values. To prevent this, the properties are validated and changed, so they are valid. This would make more sense to do in the setter functions of the properties, but the issue with that is the order matters. Eg samplerate must be at least twice as large as high-cutoff, therefore when you want to increase the high-cutoff you first have to increase the samplerate. The issue with this is that eg in gjs doing this:
```js
const cava = new AstalCava.Cava({
  samplerate: 50000,
  high_cutoff: 25000
})
``` 
would still fail, as high_cutoff would be set first (at least when i tested it).
Therefore the validation is done in the start method just before cava is initialized.

Aylur:
I tried building it on nix but I got stuck at
```
** (gjs:322174): WARNING **: 13:22:11.568: Failed to load shared library '/nix/store/kn3dy6w0gfvxs40zbl0k2mw1fqxd1jv0-astal-cava-0.1.0/lib/libastal-cava.so.0' referenced by the typelib: /nix/store/x1xy3qj2cllwrg33jyj7j28sf841cr1z-pipewire-1.2.3-jack/lib/libjack.so.0: undefined symbol: pw_log_topic_register

(gjs:322174): Gjs-CRITICAL **: 13:22:11.569: JS ERROR: GLib.Error g-invoke-error-quark: Could not locate astal_cava_get_default: /nix/store/x1xy3qj2cllwrg33jyj7j28sf841cr1z-pipewire-1.2.3-jack/lib/libjack.so.0: undefined symbol: pw_log_topic_register
```

PR #38, opened 2024-10-09
Improve showcases doc site

tokyob0t:


docs/showcases/showcases.ts
```diff
@@ -2,14 +2,23 @@ type Showcase = {
...
-    // add more showcases here
+    {
+        label: "Placeholder (this is an ags v1 screenshot)",
+        src: "/astal/showcase/aylur.png",
+        url: "https://github.com/Aylur/dotfiles",
+        icon: "devicon-javascript-plain",
+    },
+    {
+        label: "Idk I just love oxocarbon",
+        src: "/astal/showcase/contrib1.webp",
```
Aylur: please name the file something else, for example your nickname. anything thats not this generic

examples/lua/simple-bar/widget/Bar.lua
Aylur, line 1: should've been a separate PR, but it can stay now

Aylur:
I like the hover transform effect, but not on the image as it cuts of content. I like the icon too.
I feel like the title at the bottom is a bit of a bloat

I saw this showcase component on <https://its-darsh.github.io/fabric-wiki/> which I think would look cool, do you mind implementing something similar? The title and description should be optional, the author mandatory

PR #41, opened 2024-10-11
apps: better fuzzy search algorithm

kotontrion:
5The current fuzzy search algorithm just uses the levenstein distance, which does not give good results. I implemented a simple algorithms which tries to improve on that:

# results:
## before:
```bash
$ astal-apps -s fire
fish.desktop
firefox.desktop
org.gnome.Nautilus.desktop
org.gnome.Software.desktop
org.gnome.Totem.desktop
bitwarden.desktop
org.gnome.Builder.desktop
gparted.desktop
libreoffice-startcenter.desktop
com.github.kotontrion.wirecontrol.desktop
vim.desktop
```
## after:
```bash
$astal-apps -s fire
firefox.desktop
firefox-developer-edition.desktop
org.gnome.FileRoller.desktop
libreoffice-impress.desktop
enlightenment_fprint.desktop
libreoffice-writer.desktop
```

# performance:
## before
```bash
$ hyperfine "astal-apps -s fire"
Benchmark 1: astal-apps -s fire
  Time (mean ± σ):      64.7 ms ±   2.6 ms    [User: 46.8 ms, System: 16.9 ms]
  Range (min … max):    60.7 ms …  75.5 ms    47 runs
```
## after
```bash
$ hyperfine "astal-apps -s fire"
Benchmark 1: astal-apps -s fire
  Time (mean ± σ):      67.1 ms ±   2.6 ms    [User: 50.9 ms, System: 15.5 ms]
  Range (min … max):    64.4 ms …  78.3 ms    44 runs
```

Because the scoring algorithms has changed, the min_score variable, which is used to hide results, needs to be adjusted. I did not do that yet.

### DIFF
--- a/lib/apps/fuzzy.vala
+++ b/lib/apps/fuzzy.vala
@@ -1,75 +1,73 @@
 
 namespace AstalApps {
 
-    private int max(int a, int b) {
-      return a > b ? a : b;
+private int max(int a, int b) {
+    return a > b ? a : b;
+}
+
+public int fuzzy_match_string(string pattern, string str) {
+    const int unmatched_letter_penalty = -1;
+    int score = 100;
+
+    if (pattern.length == 0) return score;
+    if (str.length < pattern.length) return int.MIN;
+
+    score += unmatched_letter_penalty * (str.length - pattern.length);
+    score = fuzzy_match_recurse(pattern, str, score, true);
+
+    return score;
+}
+
+private int fuzzy_match_recurse(string pattern, string str, int score, bool first_char) {
+    if (pattern.length == 0) return score;
+
+    int match_idx = 0;
+    int offset = 0;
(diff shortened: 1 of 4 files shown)


lib/apps/fuzzy.vala
Aylur, line 1: I'll get to setup uncrustify at some point for real:D, until then I'll just ask for the styling.
move the functions to the same indent as the namespace, so
```vala
namespace AstalApps {
private int max() {

}
}
```

kotontrion:
I forgot to mention, this implementation is based on [this](https://www.forrestthewoods.com/blog/reverse_engineering_sublime_texts_fuzzy_match) with very minor modification. This algorithm provides good enough results while being very easy to understand.
The score is negative if the pattern does not match and positive otherwise. Therefore I've set the min_score to 0.

PR #42, opened 2024-10-13
showcase: kotontrion

kotontrion:
add my bar to the showcase section.


Aylur:
We definitely need some better layouts, I would love to see a long page full of these components like your menu.
But until then I'll be judging submissions. So, I would request that you open a not so noisy app on the right so the wallpaper doesn't take up most of the space. Maybe a terminal with some ascii art or even a fetch program, even though I'm not a fan of those, or maybe a file manager, whatever fits the color scheme

kotontrion:
done.

If someones interested, the music player is amberol and the audio control center is wirecontrol, which i initially wrote to test the wireplumber astal lib and then grew into its own thing.

PR #43, opened 2024-10-13
core: lua compat 5.1/5.4/luajit

tokyob0t:
`table.unpack` and `unpack` are the same function, but both are accessible in different ways in Lua 5.1/5.4/luajit. In this commit, the function unpack is defined to work in those versions without causing an error


Aylur:
you forgot about https://github.com/Aylur/astal/blob/18df91b128f6ca8850e39cb53b8536a8207e1741/core/lua/astal/variable.lua#L263

also, to not duplicate code, it would be better to polyfill table.unpack in init.lua
```lua
if not table.unpack then
    function table.unpack(tbl, from, to)
    end
end
```

tokyob0t:
oh shit i forgor to make another pr for the lua example, ignore that :[

PR #46, opened 2024-10-15
Small refactor

Aylur:
Separated out the core library into an astal-io and astal3 package

astal-io provides the glib shortcut functions such as file reading/writing/monitoring, process running, timeout functions. It might be also beneficial to implement shell like functions for example mkdir and ls.

astal3 as it name suggests contains the gtk3 widgets, most importantly the application and window classes.
astal4 is not yet implemented but it will follow the same scheme: application, window and other widgets

The gjs and lua packages have also been refactored. Their astal package no longer contains the widget and app exports they have to be imported from astal/gtk3. This makes it possible to have a single distributed package for both gtk3 and gtk4 versions.

While I was at it, after some feedback I also updated the wiki to recommend installing to /usr prefix by default and made the distinction between Astal and AGS a bit more clear

TODO:
- ~~[ ] update ags v2~~
- [x] aur packages for astal3 and astal-io


docs/guide/getting-started/installation.md
```diff
@@ -10,6 +10,8 @@ Read more about it on the [nix page](./nix#astal)
 
 maintainer: [@kotontrion](https://github.com/kotontrion)
 
+<!--TODO: fix aur package names-->
```
Aylur: @kotontrion How do you plan to name the AUR packages? The Lua package might need an update since it was moved to lang/lua from core/lua. The gjs package was also moved out of the core meson build, so an astal-js package might be beneficial when packaging ags v2 since it will depend on it.

kotontrion: the io package will be named `libastal-io` in accordance to the other libs. For the widgets one i'm not sure yet, haven't gave it any thought, but I'm leaning to reusing `libastal` for the gtk3 one and the gtk 4 version can then be named `libastal-gtk4`. (All of them will have the -git suffix, as there is no release yet).

I will update the packages after this was merged. I can fill this in the names in the docs when i have done so.

PR #49, opened 2024-10-17
core: smol python refactor

tokyob0t:


Aylur:
I'll close this, because the python lib is just a veeery experimental thing, I don't know python, I don't know how I want its API to look like and I've put it on hold until the first release of Astal.

If you are interested though, feel free to fork and do your own python thing in the meantime and open up a draft PR when you are happy with it, after the release I'll happily look at it.

PR #50, opened 2024-10-19
docs: basic lua docs

tokyob0t:


Aylur:
very nice, thanks for the help

quick question: what is the consensus on dropping the parenthesis when possible? I've never really seen any Lua codebase, this is the first Lua code I've written. I like how calling a function with a single table as the parameter does not require a parenthesis.

```lua
Widget.Box {
  Widget.Label {
    label = "",
  },
  Widget.Button {
    on_click = function() end,
  },
}
```

I like it, it makes it look like a templating language, but I've seen that for example stylua defaults to putting the parens back which is why I stuck to that for the simple bar example. What does the majority of Lua community think of this? for example awesomewm people?

tokyob0t:
there isn't an absolute consensus, rather it's more a matter of style. in the example you mentioned, it's perfectly suited for this kind of syntax and completely valid. stylua formats it that way to maintain consistency within a project, while individual users can still follow the coding styles they prefer. maybe in the case of the examples provided, that style can be followed

it's perfectly suited for this kind of syntax, just like in lua you can also define functions using `local function my_function() end` or `local my_function = function() end`, which is another example of stylistic choice. of course, the first form tends to take precedence in project standards

PR #56, opened 2024-10-24
apps: add option to include keywords in queries

matteo4375:
Added option to include keywords in queries.
Eg searching browser or email and getting the correct program.


Aylur:
* changed the default multiplier to 0.5, as the most common usecase is searching for the name of an app not for keyword and for example if another app has the name of the seached app as a keyword it does not make sense for it to have higher prio, 1000 would be way too high anyway
* realistically only a single word will be searched for in which case it does not make sense to match for mulitiple keywords, the highest matching one should be enough

PR #58, opened 2024-10-25
hyprland: fix fullscreen enum

### LINKED ISSUE #57: FULLSCREEN and MAXIMIZED are swapped in Hyprland
**Describe the bug**

In the Hyprland lib, the `Fullscreen` enum has its `MAXIMIZED` and `FULLSCREEN` values swapped. They should be 1 and 2, respectively. Additionally, there is a value missing for when a window is both maximized and fullscreen (which seems weird to me but is allowed by hyprland).

**Additional context**

Offending code: https://github.com/Aylur/astal/blob/8306ec0789854f9e04bc1708c4a7dc2afc1b4c90/lib/hyprland/client.vala#L76-L81

Relevant code in Hyprland: https://github.com/hyprwm/Hyprland/blob/f0e023bff2f2a25ffe5ed3166f55f7274d17c6bc/src/desktop/Workspace.hpp#L9-L14

Relevant Hyprland wiki entry: https://wiki.hyprland.org/Configuring/Dispatchers/#list-of-dispatchers, List of Dispatchers > fullscreenstate:
> -1 - current, 0 - none, 1 - maximize, 2 - fullscreen, 3 - maximize and fullscreen

ttoino:
Closes #57.

Swaps order of `FULLSCREEN` and `MAXIMIZED`, as per the Hyprland wiki. 

Also adds the missing value `MAXIMIZED_AND_FULLSCREEN`. This value is simply called `MAX` in Hyprland, but I thought this name would be more descriptive, suggestions welcome.

### DIFF
--- a/lib/hyprland/client.vala
+++ b/lib/hyprland/client.vala
@@ -73,10 +73,11 @@ public class Client : Object {
     }
 }
 
+[Flags]
 public enum Fullscreen {
     CURRENT = -1,
     NONE = 0,
-    FULLSCREEN = 1,
-    MAXIMIZED = 2,
+    MAXIMIZED = 1,
+    FULLSCREEN = 2,
 }
 }


Aylur:
Hyprland lib doc is in progress, and will also mention this, but we don't need the `MAXIMIZED_AND_FULLSCREEN`.

```js
if (client.fullscreen & FULLSCREEN)
  // full screen and might be maximized

if (client.fullscreen == FULLSCREEN)
  // full screen but not maximized

if (client.fullscreen & MAXIMIZED)
  // maximized and might be fullscreen

if (client.fullscreen == MAXIMIZED)
  // maximized but not fullscreen

if (client.fullscreen & (MAXIMIZED | FULLSCREEN))
  // maximized and fullscreen
```

if at any later point we implement setters for them, same thing applies

```js
client.fullscreen = FULLSCREEN // fullscreen, not maximized
client.fullscreen = MAXIMIZED // maximized, not fullscreen
client.fullscreen = MAXIMIZED | FULLSCREEN // maximized and fullscreen
```

ttoino:
That makes sense! I didn't think of bitwise operations. Although I do get this error when the last value is not defined and a window is fullscreen and maximized

```
(gjs:1261967): Gjs-CRITICAL **: 15:13:41.432: JS ERROR: Error: 3 is not a valid value for enumeration Fullscreen
```

Aylur:
When exactly do you get this error?

ttoino:
If I have some code like `bind(hyprland.focusedClient, "fullscreen").as(/* Do something with it */)` and I run `hyprctl dispatch fullscreenstate 3`. (Or set the fullscreen state to 3 some other way)

ttoino:
Looking at the Vala documentation, maybe we want to use the `[Flags]` decorator they talk about [here](https://docs.vala.dev/developer-guides/bindings/writing-a-vapi-manually/04-00-recognizing-vala-semantics-in-c-code/04-02-enums-and-flags.html), but I'm not sure how that would interact with the negative value for `CURRENT`.

Aylur:
I cannot reproduce it
```js
import GLib from "gi://GLib"
import Hyprland from "gi://AstalHyprland"

const c = Hyprland.get_default().focusedClient
print(c.title, c.fullscreen)

c.connect("notify::fullscreen", () => {
    print(c.fullscreen)
})

GLib.MainLoop.new(null, false).run()
```

I tried it with [WindowAnchor](https://aylur.github.io/libastal/astal3/method.Window.set_anchor.html) turns out that was annotated as an int which is why this did not come up, but the anchor did give this error back without `[Flags]`

I'm not sure why I can't repro with fullscreen, but it seems to work for be both with and without the `[Flags]` attribute.
Does `[Flags]` solve it for you?

ttoino:
This code produces the error for me, and is really close to the code in my dotfiles, where I encountered the error.

```tsx
// app.tsx
import { bind } from "astal";
import { App } from "astal/gtk3";
import Hyprland from "gi://AstalHyprland";

const hyprland = Hyprland.get_default();

const W = () => (
    <window application={App}>{
        bind(hyprland, "focusedClient").as((client) => (
            <label>{
                bind(client, "fullscreen").as((fullscreen) => {
                    print(fullscreen);
                    return fullscreen;
                })
            }</label>
        ))
    }</window>
);

App.start({
    main() {
        W();
    },
});
```

This is the exact error log I get:

```
(gjs:108937): Gjs-CRITICAL **: 01:40:45.948: JS ERROR: Error: 3 is not a valid value for enumeration Fullscreen
get@file:///run/user/1000/ags.js:82:54
subscribe/id<@file:///run/user/1000/ags.js:95:23
_init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:266:34
```

Using the enum with a value for 3 does not produce this error, using `[Flags]` also seems to work.

Aylur:
ah, it was the `get_fullscreen()` method

PR #62, opened 2024-10-28
new lib: clipboard

kotontrion:
Adds a clipboard library using the wlr-data-control protocol. As this uses a wlr protocol, only wlroots based compositors will be supported. 
Edit: The wlr-data-control has been adopted into the wayland protocols as ext-data-control, support in compositors is still low, but this lib will support it.

The clipboard mechanism included in the wayland core protocol only allows to get clipboard changes while the client has focus. As this lib will be used mostly in bars, which don't have focus most of the time, it is basically impossible to implement a clipboard history, as many selection events will be missed. This is fixed by using the wlroots protocol, which gets these events regardless of focus. For non-wlroots compositors I would recommend using `Gtk.Clipboard`, which implements the clipboard using the core wayland protocol.

TODO:
- [ ] docs
- [ ] support ext-data-control
- [ ] support a persistent storage backend like eg sqlite as an opt in
- [ ] add more convenience methods for the data management
- [ ] allow the user to put data into the clipboard

This PR depends on #371


rvveber:
I'm anticipating to see this

lib/clipboard/src/clipboard.c
```diff
@@ -0,0 +1,383 @@
...
+G_DEFINE_TYPE_WITH_PRIVATE(AstalClipboardClipboard, astal_clipboard_clipboard, G_TYPE_OBJECT)
+
+typedef enum {
+    ASTAL_CLIPBOARD_CLIPBOARD_PROP_SELECTION = 1,
+    ASTAL_CLIPBOARD_CLIPBOARD_PROP_PRIMARY_SELECTION,
+    ASTAL_CLIPBOARD_CLIPBOARD_PROP_MAX_LENGTH,
+    ASTAL_CLIPBOARD_CLIPBOARD_PROP_SELECTION_HISTORY,
+    ASTAL_CLIPBOARD_CLIPBOARD_PROP_PRIMARY_SELECTION_HISTORY,
+    ASTAL_CLIPBOARD_CLIPBOARD_N_PROPERTIES
+} AstalCiplboardClipboardProperties;
```
aquifolly: is this supposed to be `Ciplboard`

sadorowo:
What's the status of this PR? I see that it's abandoned for a few months. The Clipboard library will be very helpful for some of us, and we won't have to rely on `exec`/`execAsync` and some tools like `cliphist`. I'm always open if any help/testing is needed.

kotontrion:
I wasn't able to work much on astal lately, as I got occupied with other stuff and improving existing libs is more important that adding new ones IMHO, so my time went there.

This was just a proof of concept. I definitely want to revive this PR. But its gonna be a complete rewrite, there are some flaws with the approach in this PR.
 
I did some work on redesigning how Wayland protocols should be implemented in astal libs. [wl-vapi-gen](https://github.com/kotontrion/wl-vapi-gen) and #371 are key here. So before revisiting this I want to finish the astal-wl PR first and maybe implement a simpler protocol first, eg foreign-toplevel.

PR #63, opened 2024-10-29
Updated Documentation.

ghost:
Fixed a minor mistake: it's 'vala' and 'valadoc', not 'valac' for fedora.


Aylur:
can we also remove `gcc`? since it seems to be a dependency of vala anyway

ghost:
Yes, we can certainly proceed with removing `gcc` as well.

PR #65, opened 2024-11-04
core: smol lua fixes

tokyob0t:
Small fixes in Lua, the `@class` StartConfig was changed to `@alias`, and some functions were annotated


lang/lua/astal/gtk3/app.lua
```diff
@@ -24,30 +24,28 @@ end
...
----@field instance_name? string
----@field gtk_theme? string
----@field icon_theme? string
----@field cursor_theme? string
----@field css? string
----@field hold? boolean
----@field request_handler? fun(msg: string, response: fun(res: any))
----@field main? fun(...): unknown
----@field client? fun(message: fun(msg: string): string, ...): unknown
+--- @alias StartConfig {
```
Aylur: should not this be `---@alias` without a space inbetween?
also, I'm getting a lot "Fields cannot be injected in to the reference of StartConfig" with an alias

tokyob0t: yeah, it should be without the space, also for some reason, Zed works with multiline annotations, while lua-language-server doesnt

image

PR #68, opened 2024-11-07
tray: gtk4 compatibility

kotontrion:
As the core lib has started to support gtk4, the tray lib should as well. Currently it depends on gtk3 for the menus, using dbusmenu-gtk3 and for the icon lookup in themes.

- [x] menus
- [x] icon lookup
- [x] add action group as property to astalify
- [x] update examples

so something like this can be used for the tray in the future:
```js

const MenuButton = astalify(Gtk.MenuButton) //maybe support menubuttons directly, as they are useful here

function TrayItem(item) {
            return <MenuButton
                tooltipMarkup={bind(item, "tooltipMarkup")}
                usePopover={false} //only needed  for gtk3, as gtk3 popovers are broken
                actionGroup={bind(item, "action-group")} // this won't work, see below
                menuModel={bind(item, "menu-model")} >
                <icon gIcon={bind(item, "gicon")} />
            </MenuButton>
}
```

The user has to set the action group for the widget themself, I don't see a way around this. Currently you need to do this:
```
//set actiongroup ince when created and the update it on change
widget.insert_action_group("dbusmenu", item.action_group)
item.connect("notify::action-group", () => {
  widget.insert_action_group("dbusmenu", item.action_group)
})
```
It would be way more convenient to have action-group property. Maybe for the gjs bindings add the action-group property as an js object with a prefix and action-group fields, so you could do this:

```js

function TrayItem(item) {
            return <MenuButton
                tooltipMarkup={bind(item, "tooltipMarkup")}
                usePopover={false} //only needed  for gtk3, as gtk3 popovers are broken
                actionGroup={bind(item, "action-group").as(ag => return { prefix: "dbusmenu", actionGroup: ag})}
                menuModel={bind(item, "menu-model")} >
                <icon gIcon={bind(item, "gicon")} />
            </MenuButton>
}
```
Which is way better, but I don't really like it either.

Also relevant:
- [blueman menu does not work](https://gitlab.com/vala-panel-project/vala-panel-appmenu/-/issues/394)


julihermes:
Is the tray works with this merge? 
If yes, can you put a example of a code working?
I couldn't make it work.

examples/js/simple-bar/widget/Bar.tsx
```diff
@@ -11,22 +11,16 @@ import Tray from "gi://AstalTray"
...
+    return <box className="SysTray">
+        {bind(tray, "items").as(items => items.map(item => (
+            <menubutton
                 tooltipMarkup={bind(item, "tooltipMarkup")}
-                onDestroy={() => menu?.destroy()}
-                onClickRelease={self => {
-                    menu?.popup_at_widget(self, Gdk.Gravity.SOUTH, Gdk.Gravity.NORTH, null)
-                }}>
+                usePopover={false}
+                actionGroup={bind(item, "action-group").as(ag => ["dbusmenu", ag])}
```
danielwerg, line 19: `"action-group"` dash instead of underscore

examples/js/simple-bar/widget/Bar.tsx
```diff
@@ -11,22 +11,16 @@ import Tray from "gi://AstalTray"
...
+                menuModel={bind(item, "menu-model")}>
```
danielwerg, line 20: `"menu-model"` dash instead of underscore

examples/js/simple-bar/widget/Bar.tsx
danielwerg:
```
Property 'actionGroup' does not exist on type 'MenuButtonProps'.ts(2322)
```

danielwerg (commented):
had a couple issues after upgrading

examples/js/simple-bar/widget/Bar.tsx
danielwerg: https://github.com/Aylur/astal/pull/208

PR #70, opened 2024-11-10
new lib: niri

### LINKED ISSUE #8: niri ipc support
https://github.com/YaLTeR/niri/tree/main/niri-ipc
https://github.com/YaLTeR/niri/wiki/IPC

especially `EventStream` which is available in 0.1.9 (latest release)

sameoldlab:
Adds support for niri's ipc (closes #8). I took the library started by @spazzylemons, included event values in the signals, and added active and focused window/workspace tracking as properties. This covers all current messages in the event-stream but still planning to add:
- [x] output tracking (would need help testing this, as I only have one screen)
- [x] get windows by workspace
- [x] workspaces by output
- [x] ability to send actions to niri (**Done!** Fully type safe action functions with a lot of help from @maxbol)
- [x] add objects as properties instead of just their ids (**Done!** Thanks to @viceav for most of the changes)


zoenetic:
happy to help test output tracking once you get there; have niri + multiple monitors.

phaer:
@sameoldlab Great work, thank you! Are you planning to continue working on this? I think a read-only view of windows (and maybe workspaces) would already be very useful to create some bars, even if actions and state management were still missing IMO.

sameoldlab:
@phaer thanks, yes. I could mark what I have now as ready for review if that would be useful to people. I avoided it earlier because I wasn't sure if the variable names would change, but I think future updates should be possible now without any breaking changes

sameoldlab:
Posting object issues here in case anyone with better Vala knowledge has an idea. This is how to track the current title currently:
```ts
const Current = () => {
	const niri = Niri.get_default()
	const current_title: Variable<string> = Variable(niri.get_window(niri.focused_window_id)?.title ?? '')
	niri.connect('window-focus-changed', (_, w_id /* number */) => current_title
		.set(niri.get_window(w_id)?.title ?? ''))
	return new Widget.Label({ label: current_title().as(t => t.slice(0, 150)) })
}

```
Ideally it could be something similar to:
```ts
const Current = () => {
	const niri = Niri.get_default()
	return new Widget.Label({ label: niri.focused_window.title() })
}
```
While testing I still needed a variable and signal to track `changed` on `niri.focused_window` and things would break if it was ever null. The best I could figure out was a "0 object" `Window {title: "", id: -1, app_id: "", workspace_id: -1, ...}`. That would never be `null` and could change to always have the same value as the actual focused, but it would not reference the same object as `niri.get_window(niri.focused_window_id)`. 

I don't think it'd be worth the extra code if using it still needed a variable and signal, but would like to know if there's anything that might work better.

bandithedoge:
How do you plan on tackling multi-monitor setups in GTK3? Niri identifies monitors by connector name (eg. `DP-1`) but `gdk_monitor_get_connector()` is only available in GDK4.

sameoldlab:
@bandithedoge need to look at this more. The [`Output`](https://github.com/Aylur/astal/pull/70/files#diff-b222dac9a48addf308c620a8ebd51a5a16a32fd4541901982961a91ab19ac8c7) object has additional info which is available in gtk3 (make, model, logical.x, logical.y). They aren't being 'saved' yet because it needs additional messages than just reading the event stream.

The full object would make comparisons possible. For most people with just `logical.x` but not predictably enough to add an extra HashMap to the library.

sea7ones:
We really need this! @Aylur, when do you plan to merge it?

viceav:
Hi, @sameoldlab i've been testing this out and  im getting a problem where if i call `AstalNiri.get_default().workspaces` i get `null` instead of `Workspace[]`.

I dont' really know what it could be to be honest, since calling `AstalNiri.get_default().get_workspaces()` works perfectly fine.

Any help is appreciated :)

sameoldlab:
@viceav I think that's (somewhat) normal for using the libraries in javascript. `.get_thing` gives you a `Thing` js object while `.thing` has a raw `GObject<Thing>` inside. 

`null` does seem a bit odd. I could take a look but generally would say you should be using the `.get_` version instead.

viceav:
@sameoldlab found this [Error: Can't convert non-null pointer to JS value ](https://aylur.github.io/astal/guide/typescript/faq#error-can-t-convert-non-null-pointer-to-js-value), 
So probably can be fixed by adding this in the [workaround](https://github.com/Aylur/astal/blob/main/lang/gjs/src/overrides.ts)
```ts
await suppress(import("gi://AstalNiri"), ({ Niri }) => {
  patch(Niri.prototype, "workspaces")
  patch(Niri.prototype, "windows")
  patch(Niri.prototype, "outputs")
})
```

Edit: It works :)

sameoldlab:
Thanks! Do you want to make a PR @viceav? I'll just copy it in a few hours if not

viceav:
@sameoldlab Thanks! I'm gonna check if there's any other property that should be added and i'll make the PR

ayles:
When when when when when?)

maxbol:
Would love to help with this any way I can (never wrote a line of Vala/GTK code but how hard could it be?).

Also: as these additional compositor libs start getting added to astal, would it make sense to write some sort of library to abstract over them? I'm currently trying to implement that on the client side in a fork of hyprpanel to make it work with niri, and it quickly becomes quite messy since the frontend code relies on properties being bindable. Made me think it would be quite a bit easier if there was an "AstalCompositor" class implemented on the vala side.

Aylur:
@maxbol 
> would it make sense to write some sort of library to abstract over them

Yes, but not in the way you are thinking of. Wayland protocols are meant to solve this, but currently there are no protocols for some common stuff. Checkout https://github.com/quickshell-mirror/shell-protocols for an initiative.

A common class like `AstalCompositor` is out of scope and Astal will not include any common modules between compositor libraries. We should have a standalone lib for each compositor and as Wayland protocols mature we add support for those and slowly deprecate the dedicated compositor libraries such as this, the one from the sway PR and AstalHyprland.

sameoldlab:
> Would love to help with this any way I can (never wrote a line of Vala/GTK code but how hard could it be?).

Any help is welcome. The [vala docs](https://docs.vala.dev/tutorials/programming-language/main.html), skipping through [this series](https://www.youtube.com/playlist?list=PLriKzYyLb28mn2lS3c5yqMHgLREi7kR9-), and looking at the other libraries are how I learned enough to get started. Going to ~try and~ get messages done this week which should make other changes easier.

> Also: as these additional compositor libs start getting added to astal, would it make sense to write some sort of library to abstract over them? 

Yes. I remember someone mentioning this in discord a while ago. Not sure if there was any progress as a result and I haven't looked into it very much either. Some of it is possible on wayland with foreign-toplevel-mangement, workspace, and virtual-keyboard protocols, But those still don't cover everything the IPC messages can do. Even without that the community could make a higher level libraries but the ones directly in astal AFAICT are close to 1:1 bindings.

kotontrion:
> slowly deprecate the dedicated compositor libraries

I don't think that will happen. Many compositors have their own additional functionality either exposed through an IPC or custom wayland protocols. This is what makes each compositor unique. 
River, for example, has a custom wayland protocol to define custom window layouts, which isn't supported by the River lib yet, also, staying with the River example, Rivers workspaces work in a fundamentally different way from many other compositors, while i'm sure they could make the ext-workspaces protocol work with it, it definitely isn't a good fit. So the compositor specific libs are here to stay, at least for some compositors with unique features.

maxbol:
@sameoldlab Can we get in touch on discord or something? I've started making some changes in my branch and want to make sure I'm not running off in the wrong direction. Find it easier to sync development work via chat than in a PR. ~~Is there an astal discord server or something I can join? (sorry if this is plainly documented somewhere, could not find it)~~

Edit: found the button

sameoldlab:
need help from someone with an extra monitor to help test output tracking https://github.com/sameoldlab/astal/pull/18. 
cc: @maxbol @zobornost

thomascft:
> need help from someone with an extra monitor to help test output tracking [sameoldlab#18](https://github.com/sameoldlab/astal/pull/18). cc: [at]maxbol [at]zobornost

It worked for me using a slightly modified version of the ags v3 template. With my laptop and external monitor connected, I get both `eDP-1` and `DP-1`. Upon disconnecting the external monitor, its updated to just say `eDP-1`.

It's not pretty, but here's the config.

```tsx
import app from "ags/gtk4/app"
import { Astal, Gtk, Gdk } from "ags/gtk4"
import AstalNiri from "gi://AstalNiri";
import { createBinding, For } from "ags";

function Outputs() {
	const niri = AstalNiri.get_default();

	const outputs = createBinding(niri, "outputs")

	return <box>
		<For each={outputs}>
		{(output) => {
			const name = createBinding(output, "name");
			return <label label={ name(name => name.toString()) } />
		}}
		</For>
	</box>
}

export default function Test(gdkmonitor: Gdk.Monitor) {
  const { TOP, LEFT, RIGHT } = Astal.WindowAnchor

  return (
    <window
      visible
      name="Bar"
      gdkmonitor={gdkmonitor}
      exclusivity={Astal.Exclusivity.EXCLUSIVE}
      anchor={TOP | LEFT | RIGHT}
      application={app}
    >
	<box>
		<Outputs />
	</box>
    </window>
  )
}
```

sameoldlab:
Heads up. Going to remove `focused_workspace_id`, `focused_window_id`, and `focused_output_name` properties soon. You can switch to using the objects directly if updating from git.

sameoldlab:
ready for review

Aylur:
Currently trying out Niri to test out this PR, few things I have to say at first glance:

- Don't use a hashtable for event handlers, a simple switch case is a lot more readable.
- You don't have to warn on unknown events, just simply ignore in the default switch case. If a future version of Niri has a new event, users can handle it using the `event` signal until we push an update.
- Have the IPC class emit the events in the form of `public signal void event(string event, Json.Node payload)` which the Niri class handles in a switch case.
  - I'm not sure if it would be convenient or not, but we could also define each event payload as a Vala class and rely on `[Json.gobject_deserialize](https://valadoc.org/json-glib-1.0/Json.gobject_deserialize.html) 
- Don't expose the event signal with a `Json` object as a parameter. The `event` signal of the Niri class should look like `void event(string event, string payload)`, so that users can handle json as they like, for example in python and js using the json module.
- I haven't read the IPC doc of Niri yet so I'm not sure if structs need to be public, but we should prefer them to be internal. For example if an action takes some arguments which are structs we should instead expose its fields as positional arguments.

- The `msg` class should be merged with the Niri class imo.
- In the `msg` class avoid using delegates, its very hard follow. You should do it in an OOP style instead. Again, it might be better to rely on Json serialize and deserialize functions instead of manually doing it.
- Don't open a new socket connection on each action. Action functions shouldn't be static methods but instance methods that use the socket connection of the Niri singleton object.

Remember that the top priority is to make it easy and convinient to use from language bindings so avoid exposing structs and `Json`. If a struct has to be exposed, expose them as classes.

maxbol:
@Aylur Hi, I'm responsible for the delegates stuff (see discussion [here](https://github.com/sameoldlab/astal/pull/17) if interested). Originally we were trying to go the OOP route but ran into the issue of having to register serialization functions for each struct with glib-json at runtime - seemed burdensome and ugly. I guess the alternative if you want to go OOP is to replace any struct that will be serialized/deserialized with a GObject subclass?

Aylur:
The main issue is serialization of json right? How about using variants?

```vala
var variant = new Variant.array(new VariantType("{sv}"), {
    new Variant("{sv}", "key1", new Variant("s", "value")),
    new Variant("{sv}", "key2", new Variant.array(new VariantType("{sv}"), {
        new Variant("{sv}", "nested", new Variant("b", true)),
    })),
});

var serialized_json = Json.to_string(Json.gvariant_serialize(variant), true);

print(@"$serialized_json\n");
```

with a few aliases it looks quite pretty and readable

```vala
string serialized_json(Variant[] key_value_pairs, bool pretty = false) {
    var obj = new Variant.array(new VariantType("{sv}"), key_value_pairs);
    return Json.to_string(Json.gvariant_serialize(obj), pretty);
}

Variant kv_number(string key, int32 value) {
    return new Variant("{sv}", key, new Variant.int32(value));
}

Variant kv_string(string key, string value) {
    return new Variant("{sv}", key, new Variant.string(value));
}

Variant kv_bool(string key, bool value) {
    return new Variant("{sv}", key, new Variant.boolean(value));
}

Variant kv_object(string key, Variant[] key_value_pairs) {
    return new Variant("{sv}", key, new Variant.array(new VariantType("{sv}"), key_value_pairs));
}

int main (string[] args) {
    var action = serialized_json({
        kv_string("str", "string"),
        kv_bool("bool", false),
        kv_number("int", 0),
        kv_object("obj", {
            kv_string("nested", "string"),
        }),
    });

    print(@"$action\n");
    return 0;
}
```

For deserialization I would still say defining classes and using ~~[Json.gobject_deserialize](https://valadoc.org/json-glib-1.0/Json.gobject_deserialize.html)~~ [Json.gobject_from_data](https://valadoc.org/json-glib-1.0/Json.gobject_from_data.html) is the cleanest

Damianu:
"layout" property is missing on window object, it makes it impossible to properly order window list.

ItsLiyua:
Is this in a working state or still in development? Because I'd like to try it if that's possible without too much hassle

sameoldlab:
@ItsLiyua It's in a working state.
Setup is the same as the other libraries , using it will look something like this (with differences based on language):

### Get Data

Get outputs, workspaces, and windows from the default object as arrays...

```c
// psuedocode
import Niri from "AstalNiri"
niri = Niri.get_default()

// Loop through
niri.windows.forEach(w => { // do stuff })
niri.workspaces.forEach(w => { // do stuff })
niri.outputs.forEach(o => { // do stuff })
```

or even iterating through them in layers.
```c
for o in niri.outputs {
  for ws in o.workspaces {
    for win in ws.windows {
      print("window ", win.name, " on output ", ws.idx, " from output ", o.name)
    }
  }
}
```

### Send Actions
```c
// psuedocode
import Niri from "AstalNiri"
// All actions are available through Niri.msg
Niri.msg.focus_workspace_by_name("media")
// Some object specific actions can also be called through the object 
niri = Niri.get_default()
mediaWs = niri.get_workspace(3)
mediaWs.focus()
```

GTK Signals have the same names as Niri IPC events.

ItsLiyua:
Thanks! I'll try it out. When I find something that might be a bug I'll tell you

ItsLiyua:
I finally had some time to try this out and one thing I noticed was that using this pr as a nix input doesn't work because the wrapGAppsHook package was renamed to wrapGAppsHook3. I think it was fixed on the main branch but to get it to work I had to fork your repo and fix the issue myself. Currently writing a experimental bar which uses the lib. I'll report back when I tested the actual module.

ItsLiyua:
There still appears to be an issue in some way: 
```tsx
...
import Niri from "gi://AstalNiri";

export default function Bar(gdkmonitor: Gdk.Monitor) {
  const niri = Niri.get_default();
  niri.outputs.forEach((o) => {
    o.workspaces.forEach((ws) => {
      ws.windows.forEach((w) => console.log(w));
    });
  });

  return ...
}
```
This throws: 
```

(gjs:293722): ags-CRITICAL **: 00:08:29.601: CSS Error theme.scss:2:5 Expected a valid selector

(gjs:293722): Gjs-CRITICAL **: 00:08:29.825: JS ERROR: TypeError: can't access property "forEach", niri.outputs is null
Bar@file:///run/user/1000/ags.js:2004:3 -> ../../../home/liyua/nix/packages/desktop-shell/widget/bar/Bar.tsx:11:3
main@file:///run/user/1000/ags.js:2115:32
vfunc_command_line/<@file:///run/user/1000/ags.js:1437:66 -> ../../../home/liyua/nix/packages/desktop-shell/app.ts:37:24
createRoot/<@file:///run/user/1000/ags.js:869:28 -> ../../../nix/store/8irw6anf7bvrqf6nig45s270bvx2cinm-ags-js-lib/lib/gtk4/app.ts:292:17
run@file:///run/user/1000/ags.js:846:14 -> ../../../nix/store/8irw6anf7bvrqf6nig45s270bvx2cinm-ags-js-lib/node_modules/gnim/dist/jsx/scope.ts:184:28
createRoot@file:///run/user/1000/ags.js:869:16 -> ../../../nix/store/8irw6anf7bvrqf6nig45s270bvx2cinm-ags-js-lib/node_modules/gnim/dist/jsx/scope.ts:32:20
vfunc_command_line@file:///run/user/1000/ags.js:1434:17 -> ../../../nix/store/8irw6anf7bvrqf6nig45s270bvx2cinm-ags-js-lib/node_modules/gnim/dist/jsx/scope.ts:184:18
_init/GLib.MainLoop.prototype.runAsync/</<@resource:///org/gnome/gjs/modules/core/overrides/GLib.js:263:34 -> ../../../nix/store/8irw6anf7bvrqf6nig45s270bvx2cinm-ags-js-lib/lib/gtk4/app.ts:290:13
```

Might be my fault tho for using the lib incorrectly or something idk,

sameoldlab:
@ItsLiyua Sorry, the syntax I gave you earlier was just enough to explain the API. This is actually functional in javascript: 
https://github.com/sameoldlab/.dotfiles/blob/5359c8277b8781b217d4b9f4925522243c1682ef/ags/base/Bar/Workspaces.tsx#L35C1-L79C2

Please open an issue in the fork if that doesn't work (we're pinging a lot of extra people talking here).

krlade:
Any progress on this PR?

sameoldlab:
Probably not. I haven't been using ags with niri for a few months, and someone else worked on the section of ipc messaging Aylur asked for changes to.

bandithedoge:
FWIW I've been using this branch directly in [my own setup](https://codeberg.org/bandithedoge/dotfiles/src/branch/main/home/desktop/graphical/astal) for a while now with no issues. If this PR continues to be inactive, I suggest converting it to an external library rather than maintaining a fork and dealing with Git conflicts.

sameoldlab:
On one hand it's easier to find in an "abandoned" PR than a external library. But some people don't seem to like using forks. Made a copy [here](https://github.com/sameoldlab/niri-gtk). Not sure if I copied all the nix stuff correctly. Though as I said earlier I don't use it in my setup anymore, so if anyone is in a better position to maintain it please reach out.

PR #73, opened 2024-11-12
Change gir.py's to use VALADOC environment variable

### LINKED ISSUE #72: Building on Gentoo fails.
**Describe the bug**
Building on Gentoo fails with the error `FileNotFoundError: [Errno 2] No such file or directory: 'valadoc'` even though valadoc is installed.

**To Reproduce**
- Try compiling with valadoc installed via portage.

**Expected behavior**
The build to succeed.

**Additional context**
This is caused by [lib/gir.py](https://github.com/Aylur/astal/blob/2a9bb96f056f25c95b330984cfe7ac484db6c9c6/lib/gir.py), [lib/astal/io/gir.py](https://github.com/Aylur/astal/blob/2a9bb96f056f25c95b330984cfe7ac484db6c9c6/lib/astal/io/gir.py), and [lib/astal/gtk4/gir.py](https://github.com/Aylur/astal/blob/2a9bb96f056f25c95b330984cfe7ac484db6c9c6/lib/astal/gtk4/gir.py) running the `valadoc` command, but on Gentoo valadoc is installed as `valadoc-${version}`.

ghost:
This changes [lib/gir.py](https://github.com/Aylur/astal/blob/2a9bb96f056f25c95b330984cfe7ac484db6c9c6/lib/gir.py), [lib/astal/io/gir.py](https://github.com/Aylur/astal/blob/2a9bb96f056f25c95b330984cfe7ac484db6c9c6/lib/astal/io/gir.py), and [lib/astal/gtk4/gir.py](https://github.com/Aylur/astal/blob/2a9bb96f056f25c95b330984cfe7ac484db6c9c6/lib/astal/gtk4/gir.py) to check for a `VALADOC` environment variable if VALADOC is not defined, it will fall back to using `valadoc`.

Fixes #72 as `VALADOC` is defined by the [vala.eclass](https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/vala.eclass)


Aylur:
[lib/astal/io/gir.py](https://github.com/Aylur/astal/blob/2a9bb96f056f25c95b330984cfe7ac484db6c9c6/lib/astal/io/gir.py), and [lib/astal/gtk4/gir.py](https://github.com/Aylur/astal/blob/2a9bb96f056f25c95b330984cfe7ac484db6c9c6/lib/astal/gtk4/gir.py) being their own file was a not intentional, they should be removed and simply symlinked like all the other libs
Also, we should probably declare the valadoc executable as a dependency in every lib's meson.build so that if its not found it fails at the meson setup step and not in the gir script

ghost:
I'm not sure how to add the dependency in the meson.build files, so I’ll leave that to someone more familiar with meson.

PR #91, opened 2024-11-15
docs: Add example for auto-reloading css

FjedorGaede:
While migrating to the v2 I noticed that the [example from the original documentation](https://aylur.github.io/ags-docs/config/theming/#autoreload-css) on how to make auto-reloading of SCSS and CSS work was missing in the documentation. As I rely on it heavily and think it increases the DX alot - especially for new users - , it should be added to the documentation. The example I provided is the one that is running on my machine and works as expected. WDYT?

- Fjedor

### DIFF
--- a/docs/guide/typescript/theming.md
+++ b/docs/guide/typescript/theming.md
@@ -132,3 +132,53 @@ App.start({
 You could also transpile scss into css using a bundler
 and simply passing the path of the resulting css file to `css`.
 :::
+
+## Autoreloading SCSS for Improved Developer Experience
+
+To enhance the development workflow, you can add the following
+code to your `app.ts` file. This will automatically recompile
+your `style.scss` file and reload it in the app whenever
+changes are saved. With this setup, the SCSS will be recompiled
+and the app will be updated instantly on each save.
+
+Use the following code to get started:
+
+:::code-group
+
+```ts [app.ts]
+import { App } from "astal/gtk3"
+import { exec, monitorFile } from "astal"  
+
+const scss = "./style.scss" // Path to you style file
+const css = "/tmp/style.css" // Could be anywhere else
+compileScss() // Have to pre-compile the SCSS when starting up
+
+// Monitor your SCSS file for changes
+monitorFile(
+    scss,
+
+    // reload function
+    () => {
+        console.log("CSS RELOADED") // For debug purpose only
+        compileScss()
+        App.reset_css() // Have to reset the whole CSS before applying new one
+        App.apply_css(css) // Apply newly compiled CSS to the running App
+    }
+)
+
+App.start({
+    css: css,
+    main() {
+        App.get_monitors().map(Bar)
+    },
+})
+
+function compileScss() {
+    const error = exec(`sass ${scss} ${css}`) // Command to compile the SCSS file into the CSS file
+    if (error) {
+        console.error(`There went something wrong compiling the SCSS: ${error}`)
+    }
+}
+```
+
+:::


danielwerg (changes requested):
This project uses 4 spaces and doesn't use semicolons, we should keep styling consistent across entire project.

FjedorGaede:
Yes, makes sense. Added the changes.

danielwerg:
Honestly hot-reloading of both scss and typescript should be built-in ags. Interested to hear what @Aylur thinks.

For now I use `fd .* -E @girs | entr -crz ags run` until I find better but still simple tool. Normally I would use vite in frontend project.

Aylur:
I'll probably write a roadmap at some point, because this is a feature I have in mind.
For `ags run` I'm thinking of a `--watch` flag, but I don't think its possible to do hot module reloading like on web frameworks, at least I have no clue how it could be done, so all it would do is just restart the gjs process.
However for stylesheets it is possible. I'm thinking of doing it through dbus so all consumers of `Astal.Application` can benefit from it.

Aylur:
We can merge this in the meantime though after https://github.com/Aylur/astal/issues/95 is fixed

docs/guide/typescript/theming.md
```diff
@@ -132,3 +132,50 @@ App.start({
 You could also transpile scss into css using a bundler
 and simply passing the path of the resulting css file to `css`.
 :::
+
+## Autoreloading SCSS for Improved Developer Experience
+
+To enhance the development workflow, you can add the following
+code to your app.ts file. This will automatically recompile
```
danielwerg: Put file names (ex: `app.ts`) inside "`"

docs/guide/typescript/theming.md
```diff
@@ -132,3 +132,50 @@ App.start({
...
+your style.scss file and reload it in the app whenever
```
danielwerg: here too

docs/guide/typescript/theming.md
```diff
@@ -132,3 +132,50 @@ App.start({
...
+changes are saved. With this setup, the SCSS will be recompiled
+and the app will be updated instantly on each save.
+
+Use the following code to get started:
+
+:::code-group
+
+```ts [app.ts]
+const scss = `./style.scss` // Path to you style file
```
danielwerg: don't use template literals unless required (backticks)

docs/guide/typescript/theming.md
```diff
@@ -132,3 +132,50 @@ App.start({
...
+code to your app.ts file. This will automatically recompile
+your style.scss file and reload it in the app whenever
+changes are saved. With this setup, the SCSS will be recompiled
+and the app will be updated instantly on each save.
+
+Use the following code to get started:
+
+:::code-group
+
+```ts [app.ts]
```
danielwerg, line 148: add `import { exec, monitorFile } from "astal"`

docs/guide/typescript/theming.md
```diff
@@ -132,3 +132,50 @@ App.start({
...
+const css = "/tmp/style.css" // Could be anywhere else
+compileScss() // Have to pre-compile the SCSS when starting up
+
+// Monitor your SCSS file for changes
+monitorFile(
+    scss,
+
+    // reload function
+    () => {
+        console.log("CSS RELOADED") // For debug purpose only
```
danielwerg, line 159: use `print()` instead

docs/guide/typescript/theming.md
```diff
@@ -132,3 +132,50 @@ App.start({
...
+    css: css,
+    main() {
+        App.get_monitors().map(TopBar)
+    },
+})
+
+function compileScss() {
+    const result = exec(`sass ${scss} ${css}`) // Command to compile the SCSS file into the CSS file
+    if (result) {
+        console.error(`There went something wrong compiling the SCSS: ${result}`)
```
danielwerg, line 176: use `printerr()` instead

docs/guide/typescript/theming.md
```diff
@@ -132,3 +132,50 @@ App.start({
...
+App.start({
+    css: css,
+    main() {
+        App.get_monitors().map(TopBar)
+    },
+})
+
+function compileScss() {
+    const result = exec(`sass ${scss} ${css}`) // Command to compile the SCSS file into the CSS file
+    if (result) {
```
danielwerg: Maybe rename to `output` or `error`?

danielwerg (changes requested):
few suggestions

docs/guide/typescript/theming.md
danielwerg:
```ts
const scss = `${SRC}/style.scss` 
```

`SRC` could be used

FjedorGaede: Is `${SRC}` the new built-in equivalent of `${App.configDir}` or is it just something the user has to set? Because I couldn't find this anywhere in the docs and thus just used `./`.

ghost: EDIT: As Aylur says below, it's AGS-dependent, so it shouldn't be used here.

> AGS also defines a global `SRC` variable which will be replaced by the value of the --src flag. By default it will point to the directory of `entryfile`.

Source: [Bundling projects(ags docs)](https://aylur.github.io/ags/guide/bundling.html#bundling-projects)
> Since Astal is a library it lost App.configDir.
[...]
In cases where you still have to refer to the source dir, AGS defines a global SRC variable.

Source: [release notes for ags v2](https://github.com/Aylur/ags/releases/tag/v2.0.0#:~:text=In%20cases%20where%20you%20still%20have%20to%20refer%20to%20the%20source%20dir%2C%20AGS%20defines%20a%20global%20SRC%20variable.)
Also (kind of) relevant:  [setting up a project(ags docs)](https://aylur.github.io/ags/guide/init.html#setting-up-a-project:~:text=The%20env.d.ts%20will%20tell%20the%20LSP%20that%20.css%2C%20.scss%20and%20.blp%20files%20can%20be%20imported%20and%20will%20be%20inlined%20as%20a%20string.%20It%20also%20tells%20it%20that%20imports%20prefixed%20with%20inline%3A%20will%20be%20inlined%20as%20well%20as%20that%20a%20global%20SRC%20variable%20is%20available.)

Aylur: Astal docs should not use AGS features as it is meant to be independent. It is a library, there is no concept of a src directory, it's the responsibility of build tools to handle that. `./` is fine for the example, it should just have a note under it like the [one here](https://aylur.github.io/astal/guide/typescript/theming#from-file-at-startup)

FjedorGaede:
@danielwerg Good points. I pushed the changes

Aylur:
closing this because of https://github.com/Aylur/astal/pull/369

if you still feel like contributing, you can reopen this at https://github.com/Aylur/ags/

Issue #98, opened 2024-11-17
Black screen place AppLauncher


ReStranger:
**Describe the bug**
![изображение](https://github.com/user-attachments/assets/7927bbcb-3494-459e-b086-9f8754069a4d)

**To Reproduce**
I use AppLauncher from examples
I use nixos with nixpkgs-unstable

Aylur:
Please elaborate a bit. How did you run it? Any errors on terminal?

zach-martinez:
I'm having same issue. Just black screen. 
In app.ts:
```ts
App.start({
  css: style,
  main() {
    App.get_monitors().map(Bar)
    Applauncher()
  },
})
```
errors in terminal:

(gjs:77590): GLib-CRITICAL **: 22:51:25.040: g_file_test: assertion 'filename != NULL' failed

** (gjs:77590): CRITICAL **: 22:51:25.040: astal_icon_lookup_icon: assertion 'icon != NULL' failed

** (gjs:77590): WARNING **: 22:51:25.040: cannot assign (null) as icon, it is not a file nor a named icon

ReStranger:
> Please elaborate a bit. How did you run it? Any errors on terminal?

I haven't any errors in terminal
```
❯ ags run

(gjs:22076): Gtk-WARNING **: 10:34:58.209: Unknown key Settings in /home/restranger/.config/gtk-3.0/settings.ini

```

ReStranger:
![изображение](https://github.com/user-attachments/assets/cde843f6-7193-4c20-9df0-5f6041e33349)

ReStranger:
Issue in
```
<eventbox widthRequest={4000} expand onClick={hide} />
```
This create full screen box

Without `eventbox`
![Uploading изображение.png…]()

If i change  widthRequest to 100
![Uploading изображение.png…]()

ReStranger:
I set `eventbox heightRequest` to 100 and this help!!

code:
```
import Apps from "gi://AstalApps";
import { App, Astal, Gdk, Gtk } from "astal/gtk3";
import { Variable } from "astal";

const MAX_ITEMS = 8;
function hide() {
  App.get_window("AppLauncher")!.hide();
}

function AppButton({ app }: { app: Apps.Application }) {
  return (
    <button
      className="AppButton"
      onClicked={() => {
        hide();
        app.launch();
      }}
    >
      <box>
        <icon icon={app.iconName} />
        <box valign={Gtk.Align.CENTER} vertical>
[...]

PR #112, opened 2024-11-18
Fix black srceen error in applauncher example (issue #98)

ReStranger:
- Fix issue #98

### DIFF
--- a/examples/js/applauncher/widget/Applauncher.tsx
+++ b/examples/js/applauncher/widget/Applauncher.tsx
@@ -55,7 +55,7 @@ export default function Applauncher() {
                 self.hide()
         }}>
         <box>
-            <eventbox widthRequest={4000} expand onClick={hide} />
+            <eventbox widthRequest={400} expand onClick={hide} />
             <box hexpand={false} vertical>
                 <eventbox heightRequest={100} onClick={hide} />
                 <box widthRequest={500} className="Applauncher" vertical>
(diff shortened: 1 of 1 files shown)


Aylur:
please don't change the styling

ReStranger:
sorry, it's my formater

Aylur:
It is 4000 thousand because in my experience it can be any number and will still work as expected so I just chose a very high number so that it works on wide monitors too.
But if it can cause issue it should be calculated from the width of the monitor instead, or drop the this approach and just make it fullscreen.
The reason the window itself is only anchored to the top and bottom instead of all corners (making it fullscreen) is because on a fullscreen widget labels will span across the full width of the widget and you'd have to set `wrap maxWidthChars={/*number*/}` in order to wrap them, while with this current approach it setting a min-width on the window will essentially work as man-width that labels will obey and wrap automaically.

What compositor are you running on where it results in a black screen?

calebstewart:
I am also seeing this behavior (commenting here in lieu of the closed issue). I'm using Hyprland. I have three 4K monitors connected w/ resolutions of `3840x2160` with `1.25x` fractional scaling enabled. Hyprland version output, if it's helpful:

```
❯ Hyprland --version
Hyprland 0.45.0 built from branch main at commit a425fbebe4cf4238e48a42f724ef2208959d66cf  (version: bump to 0.45.0).
Date: 2024-11-09
Tag: v0.45.0, commits: a425fbebe4cf4238e48a42f724ef2208959d66cf
built against aquamarine 0.4.4


flags set:
debug
```

Changing the `widthRequest` to `400` does make the example work for me as well.

Aylur:
does https://github.com/Aylur/astal/commit/9cadfed0e4e541b8a65f4083800da6bf25c965da fix it for you? @ReStranger @calebstewart

PR #125, opened 2024-11-23
add --force arg to valadoc

festeh:
Hi!

I occasionally want to pull repo changes and update some lib with 
```
meson setup --prefix /usr build
meson install -C build
```
but I get `valadoc` error on a second command: `File already exists`. One option is to nuke `build/docs` dir each time, other is to force the generation. wdyt?


Aylur:
not needed, you can use the `--wipe` flag using meson
```sh
meson setup --prefix /usr build --wipe
meson install -C build
```

PR #134, opened 2024-11-25
set mergeBindings to minimum of 2 values and fix return type issue. 

Not-a-true-statement:
Currently if you merge any number of bindings there is a type problem where the return type can be any[] which in turn means that you can't do .as() on the return values without the hint suggestion giving a warning.

The current mergeBindings function accepts any number of elements in an array with type any as argument. In my implementation the type is restricted to bindings and has a minimum of 2 bindings that must be provided. This allows removal of filtering and cases where the length of the array is 0 or 1 since it becomes unnecessary. Also the return type is explicitly declared since the return type in the previous implementation can return a any[] (although this is unnecessary).

### DIFF
--- a/lang/gjs/src/gtk3/astalify.ts
+++ b/lang/gjs/src/gtk3/astalify.ts
@@ -6,7 +6,8 @@ import { execAsync } from "../process.js"
 import Variable from "../variable.js"
 import Binding, { kebabify, snakeify, type Connectable, type Subscribable } from "../binding.js"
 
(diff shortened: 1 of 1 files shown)


Aylur:
1. this breaks jsx children bindings
2. this function is not a public api, you should use `Variable.dervive`
3. the `any[]` parameter is intentional as it can infact take any value, which is handled in the children setter to turn everything into a widget

PR #141, opened 2024-11-26
feat: add battery_percentage to bluetooth devices

### LINKED ISSUE #138: Bluetooth Battery Levels
**Describe the solution you'd like**
I'd like to have Bluetooth.Device objects report their battery levels.

**Additional context**
I'm migrating from AGSv1, which supported battery levels on Bluetooth devices :)

ojaskavathe:
Fixes #138 

Using the Bluez [Battery API](https://github.com/bluez/bluez/blob/master/doc/org.bluez.Battery.rst) to add battery_percentage to bluetooth devices.

Can be used through:
```typescript
let mydevice = devices.get()[0]; // or map etc.
let battery = bind(mydevice, "battery_percentage");
```

### DIFF
--- a/lib/bluetooth/battery.vala
+++ b/lib/bluetooth/battery.vala
@@ -1,0 +1,33 @@
+/**
+ * Object representing a [[https://github.com/bluez/bluez/blob/master/doc/org.bluez.Battery.rst|battery]].
+ */
+public class AstalBluetooth.Battery : Object {
+    private IBattery proxy;
+
+    internal ObjectPath object_path { owned get; private set; }
+
+    internal Battery(IBattery proxy) {
+        this.proxy = proxy;
+        this.object_path = (ObjectPath)proxy.g_object_path;
+        proxy.g_properties_changed.connect((props) => {
+            var map = (HashTable<string, Variant>)props;
+            foreach (var key in map.get_keys()) {
+                var prop = kebab_case(key);
+                if (get_class().find_property(prop) != null) {
+                    notify_property(prop);
+                }
+            }
+        });
+    }
+
+    /**
+     * The percentage of battery left as an unsigned 8-bit integer.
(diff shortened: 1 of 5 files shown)


ojaskavathe:
Is anything else needed on this? @Aylur

ojaskavathe:
@Aylur still waiting on a response.

lib/bluetooth/battery.vala
```diff
@@ -0,0 +1,33 @@
+/**
+ * Object representing a [[https://github.com/bluez/bluez/blob/master/doc/org.bluez.Battery.rst|battery]].
+ */
+public class AstalBluetooth.Battery : Object {
```
Aylur: Shouldn't this be `internal`?

lib/bluetooth/device.vala
```diff
@@ -103,6 +113,16 @@ public class AstalBluetooth.Device : Object {
         set { proxy.trusted = value; }
     }
 
+    /**
+     * The percentage of battery left on the device if it has one, else 0.
+     */
+    public uint battery_percentage {
```
Aylur: I think this should be a float between 0-1. Upower also reports an int between 0 and 100 in the battery lib, but its converted into a float because most if not all consumers for example the levelbar widget uses a float between 0 and 1.
Also it should probably return -1 if there is no battery.

lib/bluetooth/battery.vala
ojaskavathe: You're right, my bad.

lib/bluetooth/device.vala
ojaskavathe: Fair enough.

andrwui:
I can't access the property on an AstalBluetooth.Device. Any guess on why that could be happening?

PR #148, opened 2024-11-29
fix(hyprland): update clients on openwindow if needed

pynappo:
Windows can open silently, which doesn't trigger activewindowv2. Therefore, we need to try adding a client on both openwindow and activewindowv2.

Wasn't sure how exactly to change the code so feel free to implement this fix differently if you'd like.

### DIFF
--- a/lib/hyprland/hyprland.vala
+++ b/lib/hyprland/hyprland.vala
@@ -303,6 +303,19 @@ public class Hyprland : Object {
         }
     }
 
+    private async bool try_add_client(string addr) throws Error {
+        if (get_client(addr) != null) {
+            return false;
+        }
+        var client = new Client();
+        _clients.insert(addr, client);
+        yield sync_clients();
+        yield sync_workspaces();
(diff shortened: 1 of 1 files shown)


ojaskavathe:
Does it make sense to notify clients on `movewindowv2` as well?

pynappo:
> Does it make sense to notify clients on `movewindowv2` as well?

I originally had that in this PR but I ended up removing it because I don't think it's needed - you can bind to a workspace's clients or a client's workspace or whatever and that should alleviate the need for an unnecessary notification.

[Here's my implementation of per-monitor taskbars, for example](https://github.com/pynappo/dotfiles/blob/5487043ec8b960ab360d07f5cb8c59a10c3cd42c/.config/ags/Bar.tsx#L316-L339)

pynappo:
@Aylur sorry for ping, but any updates? the underlying issue is preventing my taskbar from initially showing clients that open silently, like Steam.

PR #160, opened 2024-12-05
fix: typo

Ruhrozz:
it ain't much, but it's honest work


Aylur:
Its [not a typo](https://docs.gtk.org/gtk3/class.Scale.html#css-nodes)

PR #168, opened 2024-12-08
Fix simple-bar examples (special workspaces)

rRedLim:
if you use special workspace, then u will see that weird -99 workspace in ur workspaces,
you cant interact with it, and it doesnt show if it is active or no, so basically useless and ugly
2024-12-08-173713_hyprshot
2024-12-08-173818_hyprshot


rRedLim:
image
looks like special workspaces are -99...-2 ids
so now it will filter all special workspaces

rRedLim:
bruh now it definitely works xd

Aylur:
Should be also done in all three other simple bar examples with a comment explaining what id does

rRedLim:
> explaining what id does

wdym?

PR #186, opened 2024-12-17
wayland glib

### LINKED ISSUE #3: Kompass crashes on tags and wl-pointer events.
Kompass is built using `astal-river` library from this [PR](https://github.com/Aylur/astal/pull/62), meaning it conflicts with users who have the "normal" river; A quick workaround would be to apply this [patch](https://patch-diff.githubusercontent.com/raw/Aylur/astal/pull/62.patch) to Astal and then build river with it, it also adds the possibility of creating the `astal-clipboard` library itself.

kotontrion:
separates the wayland-glib integration into its own lib. This PR consists of cherrypicked commits from the #62 branch. As that will take some more time before it can get merged, I decided it would be useful to merge this change early. Especially since it also fixes a bug, which results in a freeze. 

fixes kotontrion/kompass#3


lib/wayland-glib/wl-source.vala
```diff
@@ -0,0 +1,44 @@
+namespace WlSource {
```
Aylur: Kind of dislike the namespace name
How about `namespace WlGlib` since the lib is called wayland-glib?

Aylur:
#192 will have to be done before we can merge this because the subproject symlinks outside of the river lib's root.

PR #201, opened 2024-12-23
nix: add all architectures

emsquid:
I needed `astal` to be available on all architectures for [this PR](https://github.com/Aylur/ags/pull/652)


Aylur:
Darwin is not supported, not all of the libraries can be built on darwin and the ones that can don't make much sense on darwin

emsquid:
My bad ! Then we should remove the Darwin ones and keep the aarch64-linux

PR #205, opened 2024-12-24
fix(nix): place inputs before outputs

emsquid:
inputs need to be placed before outputs otherwise I get this error
```bash
error: attribute 'aarch64-linux' missing
       at /nix/store/rdrhh93bmsz7a36xw0zi9wyn07wrlacs-source/flake.nix:47:12:
           46|         };
           47|         in astal.packages.${system} // agsPackages
             |            ^
           48|
```
switching them solve the issue.
Also please consider accepting this simple [PR](https://github.com/Aylur/ags/pull/652) to make ags available on `aarch64-linux`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,4 +1,8 @@
 {
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+  };
+
   outputs = {
     self,
     nixpkgs,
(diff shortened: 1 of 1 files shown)


emsquid:
@Aylur

Aylur:
Changing the order of an attribute set does not have any effect, and your error has nothing to do with it

PR #208, opened 2024-12-26
[Gtk3] Fix type of actionGroup prop

matt1432:
I saw that in astalify the `ActionGroup` was the one from Gtk but it should be the one from Gio so I fixed that.

Then I wanted to fix the fact that `actionGroup` was giving type errors with JSX but after looking into `Gio.ActionGroup`, I felt it was more accurate to rename the prop to `insertActionGroup` since the `actionGroup` prop isn't setting the `ActionGroup`s of the widget, it's only adding to its list which I felt was misleading.

### DIFF
--- a/lang/gjs/src/gtk3/astalify.ts
+++ b/lang/gjs/src/gtk3/astalify.ts
@@ -150,6 +150,7 @@ export type ConstructProps<
     css?: string
     cursor?: string
     clickThrough?: boolean
+    actionGroup?: ActionGroup
 }>> & Partial<{
     onDestroy: (self: Self) => unknown
     onDraw: (self: Self) => unknown
--- a/lang/gjs/src/gtk3/astalify.ts
+++ b/lang/gjs/src/gtk3/astalify.ts
@@ -3,6 +3,7 @@ import Astal from "gi://Astal?version=3.0"
 import Gtk from "gi://Gtk?version=3.0"
 import Gdk from "gi://Gdk?version=3.0"
 import GObject from "gi://GObject"
+import Gio from 'gi://Gio?version=2.0';
 import Binding, { type Connectable, type Subscribable } from "../binding.js"
 
 export { BindableProps, mergeBindings }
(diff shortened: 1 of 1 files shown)


lang/gjs/src/gtk3/astalify.ts
```diff
@@ -35,8 +36,7 @@ export default function astalify<
         get noImplicitDestroy(): boolean { return this[noImplicitDestroy] }
         set noImplicitDestroy(value: boolean) { this[noImplicitDestroy] = value }
 
-        set actionGroup([prefix, group]: ActionGroup) { this.insert_action_group(prefix, group) }
-        set_action_group(actionGroup: ActionGroup) { this.actionGroup = actionGroup }
+        set insertActionGroup([prefix, group]: ActionGroup) { this.insert_action_group(prefix, group) }
```
kotontrion: `insertActionGroup` would suggest it's a function, not a property. Keep the `actionGroup` name for the property.

It does not add the action group to a list, it replaces the action group with the given prefix in the list, so it does set the the action group of the widget with that prefix.

kotontrion (changes requested):

lang/gjs/src/gtk3/astalify.ts
matt1432: > It does not add the action group to a list, it replaces the action group with the given prefix in the list

Ahh I guess I misunderstood that part. But still, it doesn't make sense to call `actionGroup` a prop when you can have multiple groups with different prefixes and it has no getters. It is still a method that calls `insert_action_group` at the end of the day.

I don't know if it's possible, but maybe doing something like the signal props where it's `on[SomeEvent]` but it would be `actionGroupDbusmenu`? we could then actually have a getter and setter that do act like a normal property

matt1432: > I don't know if it's possible

It was not, in fact, possible.

I'll leave it like it this, makes the PR way easier now.

kotontrion (approved):

PR #220, opened 2024-12-30
[Gtk4] Extending Astal Widgets

matt1432:
This allows us to add bindable properties with correct types easily like this:

```tsx
const StyledBox = Widget.ExtendBox<{ css: string }>({
    css: (self) => {
        let _css: string | undefined;
        const provider = new Gtk.CssProvider();

        self.get_style_context().add_provider(
            provider,
            Gtk.STYLE_PROVIDER_PRIORITY_USER,
        );

        return {
            get() {
                return _css;
            },
            set(value) {
                provider.load_from_string(value);
                _css = value;
            },
        };
    },
});

(
    <StyledBox
        css={someVar().as((width) => `* { background: red; min-height: 100px; min-width: ${width}px; }`)}
    />
)
```

I am definitely open to suggestions on how to implement this, but I was pretty happy with this result.

### DIFF
--- a/lang/gjs/src/gtk4/astalify.ts
+++ b/lang/gjs/src/gtk4/astalify.ts
@@ -34,10 +34,15 @@ function _setChildren(widget: Gtk.Widget, children: any[]) {
     }
 }
 
-type Config<T extends Gtk.Widget> = {
+export type Config<T extends Gtk.Widget> = {
     setChildren(widget: T, children: any[]): void
     getChildren(widget: T): Array<Gtk.Widget>
 }
+
+export type ExtraPropsFunc<
+    Widget extends Gtk.Widget,
+    ExtraProps extends Record<string, unknown>,
+> = { [Key in keyof ExtraProps]: (self: Widget) => PropertyDescriptor }
 
 export default function astalify<
     Widget extends Gtk.Widget,
@@ -47,7 +52,7 @@ export default function astalify<
 >(
     cls: { new(...args: any[]): Widget },
     config: Partial<Config<Widget>> = {},
-    extraProps = {} as { [Key in keyof ExtraProps]: (self: Widget, values: ExtraProps) => PropertyDescriptor },
+    extraProps = {} as ExtraPropsFunc<Widget, ExtraProps>,
 ) {
     Object.assign(cls.prototype, {
         [setChildren](children: any[]) {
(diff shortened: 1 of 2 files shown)


Aylur:
There is no need for this for two reasons

1. when using jsx, you will rarely if ever need to use the getter because you should be passing stuff through props as a static value, or as `Bindings`
  ```tsx
  import { Widget, Astal, Gtk, astalify } from "astal/gtk4"
  import { Binding, Variable, bind, property, register } from "astal"
  
  interface StyledBoxProps extends Widget.BoxProps {
      css?: string | Binding<string>
  }
  
  export function StyledBox({ css, setup, ...props }: StyledBoxProps) {
      function _setup(self: Astal.Box) {
          setup?.(self)
  
          if (css) {
              // handle both `string` `Binding<string>` as a binding
              const _css = css instanceof Binding ? css : bind({
                  get: () => css,
                  subscribe: () => () => void 0,
              })
  
              self.connect("destroy", _css.subscribe(css => {
                  const provider = new Gtk.CssProvider()
                  provider.load_from_string(css)
  
                  self.get_style_context().add_provider(
                      provider,
                      Gtk.STYLE_PROVIDER_PRIORITY_USER,
                  )
              }))
          }
      }
  
      return <box setup={_setup} {...props} />
  }
  
  const css = Variable("")
  const box = <StyledBox css={css(c => `* { color: ${c}; }`)} />
  ```

2. If you really want to use getters, you should do what Gtk suggests and subclass, in this case you can subclass `Gtk.Box` otherwise subclass `Gtk.Widget` add a layout manager and only expose getters and setters you need
  ```tsx
  import { Gtk, astalify } from "astal/gtk4"
  import { Variable, property, register } from "astal"
  
  @register({ GTypeName: "StyledBox" })
  class StyledBoxClass extends Gtk.Box {
      declare private _css: string
      declare private _provider: Gtk.CssProvider
  
      @property(String)
      get css() {
          return this._css || ""
      }
  
      set css(value: string) {
          if (!this._provider) {
              this._provider = new Gtk.CssProvider()
              this.get_style_context().add_provider(
                  this._provider,
                  Gtk.STYLE_PROVIDER_PRIORITY_USER,
              )
          }
  
          this._css = value
          this._provider.load_from_string(value)
      }
  }
  
  export const StyledBox = astalify<StyledBoxClass, Gtk.Box.ConstructorProps & { css: string }>(StyledBoxClass)
  
  const css = Variable("")
  const box = <StyledBox css={css(c => `* { color: ${c}; }`)} />
  ```
  
If you feel you need to use getters a lot, I would highly suggest ditching jsx all together and just going full OOP with blueprint as a templating solution, its two different mental models

matt1432:
Thanks that makes a lot of sense. The first solution looks like it might suit my needs best.

Why do you connect to the destroy signal though?

Is the `css` prop accessible from the return value in the first case?

PR #235, opened 2025-01-05
doc(ts): explain how to have an ags-less setup

ghost:
Btw please nitpick anything you don't like

~2 things that need to be sorted out before this merge:~

~1. Excerpt from `docs/guide/getting-started/nix.md`: I couldn't get the `packages.${system}.default` derivation to work without the `--alias` line in `astalGjsBuilder`. Even adding the following wouldn't fix it: `--tsconfig="$(readlink -f $entrypoint)/tsconfig.json"`. ~
2. This is part of a bigger "problem." Excerpt from `docs/guide/typescript/init.md`: "Question for PR reviewers: Should astal-gjs always depend on Astal3?" `astal3` and `gtk3` are (especially on nix) dependencies even when not required.~

EDITS:
1. Was just a question and not very important
2. Not needed for this merge specifically

(1/24/25) I will get to removing those notes in ~2 days.
(1/28/25) Notes removed


Aylur:
closing this because of https://github.com/Aylur/astal/pull/369

AGS is kind of the built tool itself, so it would not make much sense to move this to https://github.com/Aylur/ags/
but Gnim (the js lib that was factored out from Astal) has a gtk4 template which should kind of explain how to have a AGS-less setup

PR #244, opened 2025-01-10
Added HyprPanel to showcase.

Jas-SinghFSU:


Aylur:
We will need a carousel at some point. The whole showcases tab will need some love at some point because I hope to see more contributions

PR #268, opened 2025-01-29
lua(variable): some fixes

tokyob0t:
Variable.observe now works with `notify::my-property` 👍  
Also, some functions were fixed to properly call `self.variable` methods

### DIFF
--- a/lang/lua/astal/variable.lua
+++ b/lang/lua/astal/variable.lua
@@ -33,11 +33,10 @@ function Variable.new(value)
         end
     end
 
-    variable:on_dropped(function()
+    return variable:on_dropped(function()
         variable:stop_watch()
         variable:stop_poll()
     end)
-    return variable
 end
 
 ---@param transform? fun(v: any): any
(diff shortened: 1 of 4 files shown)


lang/lua/astal/variable.lua
```diff
@@ -27,16 +27,17 @@ function Variable.new(value)
...
         variable:stop_poll()
-    end
-
-    v.on_error = function(_, err)
-        if variable.err_handler then
-            variable.err_handler(err)
-        end
-    end
+        GObject.signal_handler_disconnect(v, id)
+    end)
```
Aylur: the on_error connection does not have to be explicitly cleaned up, its done automatically
what we have now is imo more readable

lang/lua/astal/variable.lua
```diff
@@ -139,10 +139,7 @@ end
 ---@param callback function
 ---@return Variable
 function Variable:on_error(callback)
-    self.err_handler = nil
-    self.variable.on_error = function(_, err)
-        callback(err)
-    end
+    self.err_handler = callback
```
Aylur: This is not how it is meant to work. There is a default `err_handler` when `on_error` is never invoked which is just to log the error. When the user invokes `on_error` it should remove the default handler and **add** a connection, not remove previous ones-

lang/lua/astal/variable.lua
tokyob0t: understood 👍

tokyob0t:
Closing in favor of https://github.com/tokyob0t/astal

PR #269, opened 2025-01-29
 upower.vala: fix typo

cccslater:
Change `lis_...` to `lid_...`

I am doing this off my phone so I haven't checked for other occurrences !


cccslater:
Just checked -- there is one more occurrence in ifaces.vala

cccslater:
@Aylur it still says lis_... in ifaces.vala -- I didn't get around to that file and I'm presently not near a computer

PR #282, opened 2025-02-10
Add circular progress bar widget for Gtk4

ARKye03:
# Introduce a new circular progress bar widget for GTK4

> [!CAUTION]
> It's necessary to link the math library to use this widget.
>
> This explains `gtk4/src/meson.build`, new line:
>
> link_args: ['-lm']
>
> I don't use it directly, but the widget uses it under the hood.

## This widget has 4 Widgets which are:

- **Progress Arc**, the most important one:
  - `double percentage` [0..1], tells how much is drawn.
  - `int line_width` how thick the arc is
  - `Gsk.LineCap line_cap` the line cap style, Rounded, Square, Butt.
  - `Gsk.FillRule fill_rule` the fill rule, EvenOdd, Winding.
- **Radius Fill**, the radius fill area.
  - `bool radius_filled` whether the radius is filled.
  - ProgressArc props are also used to draw the radius.
- **Center Fill**, the center fill area.
  - `bool center_filled` whether the center is filled.
- **Child**, just a normal Gtk.Widget. It can be anything.

## Css nodes

```sh
circular-progress
├── progress-arc
├── center-fill
╰── radius-fill
```

Use the CSS `color` property to set the color of each node.

## Example

With Gtk Blueprint:

```blp
$CircularProgressBar {
    percentage: bind template.speaker as <AstalWp.Endpoint>.volume;
    line-width: 5;
    line-cap: round;
    center-filled: false;
    radius-filled: true;

    child: Image {
    icon-name: bind template.speaker as <AstalWp.Endpoint>.volume-icon;
    };
}
```

cpb

The icon looks unimpressive, but it's not an issue with the CircularProgressBar.


kotontrion:
I haven't looked at the widget implementation yet.
Please also add it to the jsx-runtime so you can use it in the gjs bindings using jsx. Note that in gtk4 astal makes use of Gtk.Builder to build the widget tree from jsx, so you need to inherit from Gtk.Buildable and at least implement its `add_child` method.

This would then also allow to this in blueprint instead of specifying the child prop:
```
$CircularProgressBar {
    percentage: bind template.speaker as <AstalWp.Endpoint>.volume;
    line-width: 5;
    line-cap: round;
    center-filled: false;
    radius-filled: true;

    Image {
        icon-name: bind template.speaker as <AstalWp.Endpoint>.volume-icon;
    };
}
```

ARKye03:
> This would then also allow to this in blueprint instead of specifying the child prop:

It implements `Gtk.Buildable`, and `Gtk.Buildable.add_child`:

```blp
Astal.CircularProgressBar {
  percentage: bind template.speaker as <AstalWp.Endpoint>.volume;
  line-width: 5;
  line-cap: round;
  center-filled: true;
  radius-filled: true;
  fill-rule: winding;

  Image {
    icon-name: bind template.speaker as <AstalWp.Endpoint>.volume-icon;
  }
}
```

I have added GJS Binding, which I think they're correct, suggestion/corrections would be great. Valid snippet:

```tsx
<CircularProgress
    lineCap={Gsk.LineCap.ROUND}
    lineWidth={5}
    percentage={0.5}
    centerFilled={true}
    radiusFilled={true}
    fillRule={Gsk.FillRule.EVEN_ODD}
    widthRequest={30}
><image
    iconName={"audio-volume-high-symbolic"}
/>
</CircularProgress>
```

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,521 @@
...
+    }
+
+    /*
+     * Implements Gtk.Buildable interface.
+     */
+    public void add_child(Gtk.Builder builder, GLib.Object child, string? type) {
+        if (child is Gtk.Widget) {
+            this.child = (Gtk.Widget)child;
+        } else if (child is Gtk.EventController) {
+            this.add_controller((Gtk.EventController)child);
```
kotontrion: It would be better to just chain up here using `base.add_child` instead of handling event controllers yourself

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,521 @@
...
+        }
+    }
+
+    static construct {
+        set_css_name("circular-progress");
+        Gtk.CssProvider css_provider = new Gtk.CssProvider();
+        css_provider.load_from_string(
+            """
```
kotontrion: I can see why you want to add some minimal css for this widget, but I don't really like having it in code here. What I did in libkompass is have one scss file containing some minimal definitions for all my widgets, which is loaded as a gresource when the lib is initiated, similar to what libadwaita does.

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,521 @@
...
+    public void add_child(Gtk.Builder builder, GLib.Object child, string? type) {
+        if (child is Gtk.Widget) {
+            this.child = (Gtk.Widget)child;
+        } else if (child is Gtk.EventController) {
+            this.add_controller((Gtk.EventController)child);
+        }
+    }
+
+    static construct {
+        set_css_name("circular-progress");
```
kotontrion: in gtk in general css names do not contain dashes, so this should be `circularprogress` for consistency. This also applies to the the subwidgets.

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,521 @@
...
+    }
+
+    private void invalidate_dimensions() {
+        _cached_width = -1;
+        _cached_height = -1;
+        _cached_radius = -1;
+        _cached_delta = -1;
+    }
+
+    public override void measure(Gtk.Orientation orientation,
```
kotontrion: When a widget has a layout manager set, gtk will always use the layout managers measure function over your implementation. so this function is never called and can be removed.

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,521 @@
...
+            );
+            minimum = natural = int.max(24, child_min);
+        }
+    }
+}
+
+/**
+ * Private widget that handles drawing the progress arc.
+ */
+private class ProgressArc : Gtk.Widget {
```
kotontrion: Gtk does have a private class Gtk.Gizmo, which is basically just a Gtk.Widget, but takes function pointers for snapshot, allocate, measure, etc. Reimplementing this for astal could be quite useful for other custom widgets in the future as well. Such a class would allow you to have lambdas for those functions and would get rid of the duplicated properties in each subwidget and the update_geometry function.

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,521 @@
...
+        _radius_fill.unparent();
+        _radius_fill = null;
+        base.dispose();
+    }
+
+    public override void snapshot(Gtk.Snapshot snapshot) {
+        var width = get_width();
+        var height = get_height();
+
+        // Only recalculate if dimensions changed
```
kotontrion: I don't think caching those values is really useful. The expensive part is the drawing in the snapshot, which you do anyway. Caching the gsk path would have more impact. Also smarter redraw management. Eg when the percentage changes only the progressarc needs to be redrawn, but you call update_geometry and queue_draw on all subwidgets, but center and radius fill did not change.

Generally I don't think such performance optimizations are really necessary, as the drawing calls are only a few simple ones and it in practice the properties are not changing frequently.

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,521 @@
...
+                _child.notify.connect(() => queue_draw());
+            }
+        });
+    }
+
+    public CircularProgressBar() {
+        Object(
+            name : "circular-progress"
+        );
+        notify.connect(() => {
```
kotontrion: you call queue_draw in every property setter and connect to the notify signal here to do the same thing. So when the Widget is created using this constructor, queue_draw will be called twice. Gtk will optimize that out, but still unnecessary. I would suggest removing the queue_draw calls from the setters, and move this handler to the construct function above. This way you won't need to have private variables for most properties anymore making it more readable.

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,521 @@
...
+        layout_manager = new Gtk.BinLayout();
+        overflow = Gtk.Overflow.HIDDEN;
+        can_focus = true;
+        focusable = true;
+        can_target = true;
+
+        // Connect to child's notify signal to redraw when child changes
+        notify["child"].connect(() => {
+            if (_child != null) {
+                _child.notify.connect(() => queue_draw());
```
kotontrion: I don't think that is necessary. The only time you want to redraw when the child changes is when it resulted in a size change of the progressbar. But in that case gtk will know that your allocation has changed and that you need a redraw, so this handler just results in more unnecessary redraws.

kotontrion (changes requested):
Two more things to add:
- add circularprogress to the jsx runtime in the jsx-runtime.ts file.
- fill rule and line cap properties require Gsk, so add this to the reexports in the index.ts file, so you can do
```js
//this
import { App, Astal, Gtk, Gdk, Gsk } from "astal/gtk4"
//instead of
import Gsk from "gi://Gsk"
```

Also I would like to wait for Aylurs opinion on the Css style sheet and the Gizmo changes.

ARKye03:
I will implement all or most of the changes. Nevertheless, I will put this widget on standby, as a hypothetical `Astal.Gizmo` seems more suitable to create first. Then, I will use it to develop a circular progress widget.

ARKye03:
Changes made:
- Added circularprogress to the JSX runtime in the jsx-runtime.ts file.
- Added Gsk to the reexports in the index.ts file.
- Removed cached values(radius, delta, width, height), unnecessary "optimization"
-  Removed private properties, `bool _center_filled`, `bool _radius_filled`, `Gsk.LineCap _line_cap`, `Gsk.FillRule _fill_rule`, and used public variable directly for most cases.
- Removed `queue_draw()` calls on every setter and moved to the constructor, called on each property change.
- Used base.add_child instead of handling event controllers manually.
- Changed CSS Nodes to `circularprogress`, `progress`, `radius`, and `center` to make it consistent and simpler.
- Removed unused `measure()` method as the CircularProgress uses [BinLayout](https://docs.gtk.org/gtk4/class.BinLayout.html) as LayoutManager
- Removed LayoutManager to the different nodes of CircularProgress, as these don't and won't have children.
- Changed and formatted the code style to make it subjectively more readable and consistent with the rest of the code.

> [!WARNING]
> This still doesn't use Astal.Gizmo

ARKye03:
> [!WARNING]
> Below will only happen if Astal.Gizmo or similar is merged.

I’ve created a variant of this widget using `Astal.Gizmo` from #293 , which you can find [here](https://github.com/ARKye03/astal/tree/feat/AstalCircularProgressGizmo). If the Gizmo gets approved, I’ll use that version and force merge the branch into this one.

lang/gjs/src/gtk4/jsx-runtime.ts
```diff
@@ -22,7 +22,7 @@ const ctors = {
     box: Widget.Box,
     button: Widget.Button,
     centerbox: Widget.CenterBox,
-    // circularprogress: Widget.CircularProgress,
+    circularprogress: Widget.CircularProgress,
```
WaltoSoft, line 25: I've been watching this PR hoping it makes it through.  I think you also need to add circularprogress in the IntrinsicElements interface a bit further down in this file.

lib/astal/gtk4/src/widget/circularprogress.vala
kotontrion: You moved this into the child setter, but my argument still stands. Connecting to the childs notify signal should not be necessary.

ARKye03: If I don't do this, the child might change and it won't update. Take for instance this:
```blp
Astal.CircularProgressBar {
  percentage: bind template.speaker as <AstalWp.Endpoint>.volume;
  line-width: 5;

  Gtk.GestureClick {
    button: 1;
    pressed => $toggle_volume();
  }

  Image {
    icon-name: bind template.speaker as <AstalWp.Endpoint>.volume-icon;
  }
}
```
```cs
[GtkCallback]
public void toggle_volume() {
	speaker.mute = !speaker.mute;
}
```

Clicking here will toggle the audio's mute state, meaning the icon should change. Removing `child.notify.connect` will cause the icon to remain unchanged. Showing Mute state until some other property changes, this might improve by redrawing only the child, and should be made to other nodes as well, I don't know how to make it clean and I was waiting for the Gizmo PR to dedicate more time here. So any suggestion in this regard would be great, so I can fulfill Astal's needs.

TL;DR: If the icon solely changes it won't update correctly.

ARKye03: What's left here, in my consideration, is handling the disconnection of the child and a better Dispose(that also handles the child and its notifications) I already did this, and I'll commit it if and only if no better approach is found by me or anyone.

kotontrion: Then there is a bug somewhere.
When the images icon-name changes the image has to trigger a redraw itself. You are not responsible for monitoring your child properties to check if they might need a redraw.

I looked at your snapshot function again and found the issue. You need to call snapshop_child to snapshot a child, or subwidget.

ARKye03: Well, that makes me a bit of a retard, I did know about snapshot_child(), just ignored it. 

Done, thanks for the help.

lib/astal/gtk4/src/widget/circularprogress.vala
ARKye03: I removed the minimal CSS; I didn't like it either, so better implement this later for a more generic use case.
Maybe another PR to load a custom CssProvider based on a hypothetical `astal.scss`.

lib/astal/gtk4/src/widget/circularprogress.vala
ARKye03: > [!NOTE]
> Feel free to ignore this comment, as this depends entirely on the Gizmo PR, and this might not be the time to talk about this.

[This](https://github.com/ARKye03/astal/blob/feat/AstalCircularProgressGizmo/lib/astal/gtk4/src/widget/circularprogress.vala) removes this redundancy, and no longer `update_geometry` methods exist. Nevertheless, the custom nodes (Progress, Radius, Center) are manually snapshotted, e.g., `_progress_arc.snapshot(snapshot)` instead of `snapshot_child(_progress_arc, snapshot)` which doesn't seem to work correctly, but for the `_child` it works as intended. It could and looks like a bug, but the Gizmo looks fine, and again, it's just the original `gtkgizmo.c` but in Vala.

Right now is:
```cs
if (center_filled) {
    _center_fill.snapshot(snapshot);
}

if (radius_filled) {
    _radius_fill.snapshot(snapshot);
}

_progress_arc.snapshot(snapshot);

if (_child != null) {
    snapshot_child(_child, snapshot);
}
```
But it could directly be:

```cs
if (center_filled) {
    draw_center_fill(snapshot);
}

if (radius_filled) {
    draw_radius_fill(snapshot);
}

draw_progress_arc(snapshot);

if (_child != null) {
    snapshot_child(_child, snapshot);
}
```

Calling the methods directly will allow the same logic with no extra overhead unless there is something I'm missing.

kotontrion:
I totally missed that this implementation is missing a feature which the gtk3 version has, and many people probably want. 

The gtk3 version allows to specify a start and end point, so it does not draw a full circle, but only an arc.

You could then also make a function which draws an arc, which gets used by both the radius and progress fill snapshot functions to deduplicate code.

ARKye03:
> I totally missed that this implementation is missing a feature which the gtk3 version has, and many people probably want.
> 
> The gtk3 version allows to specify a start and end point, so it does not draw a full circle, but only an arc.
> 
> You could then also make a function which draws an arc, which gets used by both the radius and progress fill snapshot functions to deduplicate code.

I have implemented the `start_at` and `end_at` properties, both works. Also implemented the `inverted` property, so now you can draw a progress bar clockwise and counter-clockwise. I have refactored the snapshot of the progressArc, to be more "readable".

Both `start_at` and `end_at` can have values from -1 to 1, where:
- `-1 => -2π`
- ` 0 => 0`
-  `1 => 2π` 

Kinda obvious. Also, there can be a lot of refactoring, like you said having a draw arc for both radius and progress, could be possible, but some issues exist when trying to draw a complete arc, which is easily avoidable by drawing a circle, and cleaner. Using the `Astal.Gizmo` offers a lot of room for refactoring too.

JulesdeCube:
Hello and first thank you for your work.

I was planning to migrate my config to gtk4 but the absence of this widget block me :sweat:.

I see that #293 is quite old. Is #293 still plan ?
If not can we merge this pr in the current state ?

NB: I totally understand that this is an opensource project without any guaranteed. this message is just to understand current status.

ARKye03:
> I see that #293 is quite old. Is #293 still plan ? If not can we merge this pr in the current state ?

I use that gizmo every day. I think it is just a simple widget that doesn't add any overhead and acts as a helper. On the other hand, I consider that this PR, in its current state, is obsolete or at least not the best approach, since I use a [variant](https://github.com/ARKye03/morghulis/blob/trunk/src/Modules/CircularProgress.vala) based on the gizmo from #293, which I believe looks cleaner. So, merging `Astal.Gizmo` would simplify this widget's structure.

ARKye03:
I have completely refactored the widget with the recent merge of #293, so once again, this widget is ready for review.

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,544 @@
...
+        }
+    }
+
+    /**
+     * The starting position (-1.0 to 1.0).
+     * - -1.0 = -2π degrees (3 o'clock)
+     * - -0.5 = π degrees (9 o'clock)
+     * - 0.0 = 0 degrees (3 o'clock)
+     * - 0.5 = π degrees (9 o'clock)
+     * - 1.0 = 2π degrees (3 o'clock)
```
kotontrion, line 104: - later in the code you always need to multiply it with 2π, so why not make the range from -2π to 2π instead of from -1 to 1?
- the table fails to specify if the angle is measured clockwise or anti-clockwise and does not render properly in the docs.
something like this would be better:
`the starting angle of the progress arc, measured in radians, ranging from -2π to 2π relative to the positive x-axis direction.`

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,544 @@
...
+        }
+    }
+
+    private void normalize_angles() {
+        if (_end_at < _start_at) {
+            var temp = _end_at;
+            _end_at = _start_at;
+            _start_at = temp;
+            inverted = !inverted;
+            angles_changed();
```
kotontrion, line 146: changing the angle properties is bad practice, it leads to unexpected behavior when the start-at/end-at properties are used in a binding, especially when you don't emit the notify signals. This should be handled in the draw functions.

lib/astal/gtk4/src/widget/circularprogress.vala
```diff
@@ -0,0 +1,544 @@
...
+                _start_at = value;
+            }
+            normalize_angles();
+        }
+    }
+
+    /**
+     * The ending position (-1.0 to 1.0).
+     *
+     * Similar to {@link start_at}, this property determines the angular position
```
kotontrion, line 123: use `[property@Astal.CircularProgressBar:start-at]` to properly link to another property.

kotontrion (changes requested):
A few things i noticed, that either changed since my last review, or I missed them, or changed my mind about (not sure, it has been a while):
- the start-at angle is measured clockwise, while the percentage is measured counter-clockwise, which is weird. Mathematical convention for angles is counter-clockwise, but it should be at least consistent within the widget.
- the way the inverted prop works is inconsistent with how it works in gtks scale widget. Here you calculate the distance between the start and end angle, then draw an arc starting at the start angle either in clockwise or counter-clockwise direction based on inverted, this results in the effective end angle changing based on the inverted prop, which is unexpected. It would be more consistent with gtks scale widget to draw clockwise from start angle to end angle when not inverted and counter-clockwise from end angle to start angle when inverted. This would then also solve the issue of the effective end angle changing with when start angle changes when not inverted.
- The radius-fill is always drawn as a full circle, but it should only draw the arc from start angle to end angle.

PR #288, opened 2025-02-16
python bindings

0x6e6174:
adds python bindings.

todo:
- [ ] docs, examples
- [x] rewrite the variable class to match the {lua,gjs} implementations instead of being bodged natalie code
- [x] improve the way gobject libraries are imported/improve general packaging structure
- [x] update set_children in astalify.py to work correctly
- [ ] nix packaging?
- [x] general code cleanup
- [ ] gtk4?


Weathercold:
I wrote a nix package https://github.com/Weathercold/cryosphere/blob/hy/nix/pkgs/as/astal-py/package.nix

0x6e6174:
@Weathercold would you mind submitting a pull request to 0x6e6174/astal with the nix package, so it can get merged when the rest of this does?

0x6e6174:
also, does that build? poetry requires python3.12 because lxml wouldn't compile on 3.13, but in the nix package, you have it under python3.13

Weathercold:
It does build. If you use it you can just change to python 3.12

Although I haven't yet managed to display a basic bar written in hy

0x6e6174:
what problems are you having with hy? you can use https://git.natalieee.net/nat/ui as an example if you haven't seen that yet.

Weathercold:
> @Weathercold would you mind submitting a pull request to 0x6e6174/astal with the nix package, so it can get merged when the rest of this does?

I'll do that once I find some free time

> what problems are you having with hy? you can use https://git.natalieee.net/nat/ui as an example if you haven't seen that yet.

The program runs but nothing is displayed https://github.com/Weathercold/cryosphere/blob/hy/src/main.hy

0x6e6174:
Ah. the issue is that map statements in python are lazy iterators. 

map looks something like this:
```py
def map(f, iter):
    for item in iter:
        yield f(item)
```

as such, when you map Bar over App.get_monitors(), Bar does not get called until you attempt to access the result of the map statement. you can fix this by either using a list comprehension, which is not lazily evaluated (lfor in hy), or by converting the map statement to a list.

Weathercold:
Thanks, I got the bar working

Btw is monitor_file working? It seems that it can't detect file save whether I monitor a file or a directory

0x6e6174:
terminal a:
```
~ (130) λ hy
Hy 1.0.0 (Afternoon Review) using CPython(main) 3.12.9 on Linux
=> (import astal *)
=> (import astal.gtk3 *)
=> (monitor-file "/tmp/foo" print)
<__gi__.GInotifyFileMonitor object at 0x7f7828801140 (GInotifyFileMonitor at 0x5635286a3dc0)>
=> (.start App)
/tmp/foo 0
/tmp/foo 1
/tmp/foo 4
```

this is the result of running `vim /tmp/foo -c ':wq'`, can you elaborate on the issue you are having?

0x6e6174:
closing this. as per a conversation had in the discord, bindings for libastal will be moved to their own repos. future development of python bindings will take place here: https://github.com/0x6e6174/astal-py.

PR #293, opened 2025-02-20
Add new Gizmo widget helper to create custom widgets in GTK4

ARKye03:
While [gtkgizmo.c](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkgizmo.c) is quite simple, I need to review the code thoroughly and test it in the field (e.g. [CircularProgress](https://github.com/ARKye03/astal/tree/feat/AstalCircularProgressGizmo)). 

> [!NOTE]
> CircularProgress above will be merged onto https://github.com/ARKye03/astal/tree/feat/CircularProgressWidget, to keep work at #282

### DIFF
--- a/lib/astal/gtk4/src/meson.build
+++ b/lib/astal/gtk4/src/meson.build
@@ -29,6 +29,7 @@ sources = [config] + files(
   'widget/box.vala',
   'widget/slider.vala',
   'widget/window.vala',
+  'widget/gizmo.vala',
   'application.vala',
 )
 
--- a/lib/astal/gtk4/src/widget/gizmo.vala
+++ b/lib/astal/gtk4/src/widget/gizmo.vala
@@ -1,3 +1,8 @@
+/**
+ * Helper widget to create custom widgets
+ *
+ * Based on private GtkGizmo widget
+ */
 public class Astal.Gizmo : Gtk.Widget {
     public delegate void MeasureFunc(
         Gtk.Orientation orientation, int for_size,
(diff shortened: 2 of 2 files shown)


ARKye03:
I used it to construct the CircularProgress keeping it more readable and simple, and it works. So, it's ready for review. This widget is no more than a ""copy-paste"" of the gtkgizmo, but in Vala.

kotontrion:
> Should AstalGizmo implement Gtk.Buildable?

No. Creating a Gizmo with a Builder is nonesense, as you can't define the functions that way. 

Also using function pointers like here in gobject introspection is not that straight forward. It would involve a bit more work. I think the best cause of action would be making that class `internal` instead of `public`, so it can be used by astal, but it's not exposed in the gir.

Otherwise looks good to me, didn't test it though.

ARKye03:
> > Should AstalGizmo implement Gtk.Buildable?
> 
> No. Creating a Gizmo with a Builder is nonesense, as you can't define the functions that way.

Yes, I realized this rather quickly.

> Also using function pointers like here in gobject introspection is not that straight forward. It would involve a bit more work. I think the best cause of action would be making that class `internal` instead of `public`, so it can be used by astal, but it's not exposed in the gir.
> 

I made it internal

Now I can rest and ignore GJS bindings <3

JulesdeCube:
following on https://github.com/Aylur/astal/pull/282#issuecomment-3065702333

Hello, from what I see this pr seems to be ready to be merged. 

Is there is any blocker ?

PR #297, opened 2025-02-21
AstalWp: add missing features

### LINKED ISSUE #295: Make AstalWp.Device give the `form-factor`
**Is your feature request related to a problem? Please describe.**
Often `device.icon` gives me too little control over what the icons look like (I have no idea what the full set of icons a device can ask for is, so it's easy for an icon i don't have to show up). My solution to this in AGS v1 was just to fallback using the form_factor instead.

**Describe the solution you'd like**
It would be nice if a AstalWp.Device contained form factor info

**Describe alternatives you've considered**
I could just live with it and have a single fallback icon (`<icon icon={Icon.lookup_icon(d.icon) ? d.icon : "audio-x-generic"}/>`) instead of using the form factor to more accurately guess.

**Additional context**
I have no idea where form factor actually comes from. It's possible it isn't a wireplumber thing, in which case feel free to completely disregard this issue, but I'm strugging to find any docs which list all the valid wireplumber device properties.

### LINKED ISSUE #318: attributes missing for enums in gir
In the gir files the enums are missing the `glib:type-name` and `glib:get-type` attributes.

These attributes (actually only the get-type) are needed for the haskell bindings, maybe other languages are affected as well, but gjs and lua seem to work fine without it.

This affects only the libs written in C, so Cava, WP and River (Auth does not have enums).

Edit: fixed for cava and river as of bedc1fe771bf30460fc3f250d5530fb0c5c30790
For the Wp lib, I will address this in the #297 PR

kotontrion:
The AstalWp lib initially only implemented enough to achieve feature parity with the audio service of ags v1 and is missing some useful features. 

## Todo:
- [x] per channel volume controls
- [x] set an endpoints target device
- [x] configure device routes
- [x] give the user access to more pipewire properties
- [x] docs

If there are feature missing in the list above, let me know.

### breaking
- the `AstalWpEndpoint` class was renamed to `AstalWpNode`. There are two new Classes: `AstalWpStream` and `AstalWpEndpoint`, both inherit from `AstalWpNode`. 
- `is-default` property was moved from AstalWpNode to AstalWpEndpoint
- use `glib-mkenums`, which might break packages. eg in Arch it is part of glib-devel, so this needs to be added as a build dependency

### non breaking
- AstalWpStream 
  - `target-serial`: `int` represents the serial number of the Endpoint this stream targets.
  - `target-endpoint`: `AstalWpEndpoint` the target endpoint of this stream
  - `media-role`
  - `media-category``
- AstalWpEndpoint
  - `device-id`: `uint` the id of the device associated with this endpoint
  - `device`: `AstalWpdevice` same, but the device object instead of id
  - `is-default`: `boolean` was moved here from AstalWpNode
  - `routes`: List of `AstalWpRoute` objects
  - `active-route-id` and `active-route` for the currently active route
- AstalWpNode
  - `state` : `AstalWpNodeState` the current state of this node (eg, running, idle)
  - `channel-volumes`: `GList(AstalWpChannelVolume)`  a list containing volume information for each channel
  - `get_pw_property`: gets a property of the underlying pipewire object. This allows to read properties, which are not wrapped as GObject properties yet.
- AstalWpdevice
  - `form-factor`: `string`
- AstalWpChanelVolume
  - `name`: `string` the name of the channel
  - `volume`: `double` the volume
  - `volume-icon`: `string` the volume icon
- AstalWpRoute object representing a Route
- AstalWpWp now has a `ready` signal, which is emitted after streams/endpoints/devices/etc are initially loaded

Note: kotontrion/wirecontrol#1 does make use of these new feature.

closes #295
closes #318


PartyWumpus:
Hey, I was bored so I decided to see if I could implement this stuff on my own. I've got form-factor and and endpoint target device on a branch [PartyWumpus/astal/wireplumber-improvements](https://github.com/PartyWumpus/astal/tree/wireplumber-improvements). No idea if what I've written is any use, but thought I'd share. I could make a PR but this does what you're already planning to do so I wasn't sure.

(sidenote: being able to instantly just use `endpoint.device.formFactor` in TS with only these quick changes is great, astal is a really cool tool)

kotontrion:
This PR is now basically ready except for the docs and a little cleanup. Also requires a bit more testing. I did only test it using wirecontrol, linked in the original message.

PR #298, opened 2025-02-22
feat(hyprland): Client groups

spotdemo4:
I'd like to be able to get whether or not a client is in a group, and what other clients are in that group. 

Wasn't sure if it was TODO with a Group[] class because you were going to return more information about each client in the group, just getting the address is fine for my use case

### DIFF
--- a/lib/hyprland/client.vala
+++ b/lib/hyprland/client.vala
@@ -23,8 +23,8 @@ public class Client : Object {
     public Fullscreen fullscreen { get; private set; }
     public Fullscreen fullscreen_client { get; private set; }
 
-    // TODO: public Group[] grouped { get; private set; }
     // TODO: public Tag[] tags { get; private set; }
+    public Array<string> grouped { get; private set; }
     public string swallowing { get; private set; }
     public int focus_history_id { get; private set; }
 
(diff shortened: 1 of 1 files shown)


Mik-813:
Hey, @spotdemo4, I also needed that, so I tried using your modification, which worked on vala, but building it and using the "grouped" property in gjs (ags) resulted in the following error `JS ERROR: Error: Unknown signal`. Have you by any chance encountered and solved that issue?

lib/hyprland/client.vala
```diff
@@ -23,8 +23,8 @@ public class Client : Object {
     public Fullscreen fullscreen { get; private set; }
     public Fullscreen fullscreen_client { get; private set; }
 
-    // TODO: public Group[] grouped { get; private set; }
     // TODO: public Tag[] tags { get; private set; }
+    public Array<string> grouped { get; private set; }
```
Mik-813, line 27:
```suggestion
    public string[] grouped { get; private set; }
```

lib/hyprland/client.vala
```diff
@@ -51,6 +51,10 @@ public class Client : Object {
         fullscreen = (Fullscreen)obj.get_int_member("fullscreen");
         fullscreen_client = (Fullscreen)obj.get_int_member("fullscreenClient");
 
+        grouped = new Array<string>();
+        foreach (var addr in obj.get_array_member("grouped").get_elements())
+            grouped.append_val(addr.get_string().replace("0x", ""));
+
```
Mik-813, line 57:
```suggestion
        string[] _grouped = {};
        foreach (unowned Json.Node node in obj.get_array_member("grouped").get_elements()) {
            unowned string raw_val = node.get_string();
            if (raw_val != null) {
                _grouped += raw_val.replace("0x", "");
            }
        }
        grouped = _grouped;
```

Mik-813 (commented):
I changed an `Array<string>` to `string[]` and now it works with js

Mik-813:
I created a pull request that addresses groups feature more thoroughly:
https://github.com/Aylur/astal/pull/428

spotdemo4:
Closing in favor of #428 (I haven't looked at this in almost a year)

PR #303, opened 2025-02-27
feat: follow symlinks in file write and monitor functions

### LINKED ISSUE #95: AstalIO.monitorFile crashes on nested directories
**Describe the bug**
[AstalIO.monitorFile](https://aylur.github.io/libastal/io/func.monitor_file.html) fails to walk through directories recursively

davfsa:
As title describes, update file utilities to respect and follow symlinks (which Vala's `FileUtils` doesn't do by default) when writing and monitoring files.

also fixes #95

### DIFF
--- a/lib/astal/io/file.vala
+++ b/lib/astal/io/file.vala
@@ -24,11 +24,17 @@ public async string read_file_async(string path) throws Error {
 /**
  * Write content to a file synchronously.
  */
-public void write_file(string path, string content) {
+public void write_file(string path, string content, bool follow_symlinks = true) {
     try {
         var dir = Path.get_dirname(path);
         if (!FileUtils.test(dir, FileTest.IS_DIR)) {
             File.new_for_path(dir).make_directory_with_parents(null);
+        }
+
+        if (follow_symlinks) {
+            while (FileUtils.test(path, FileTest.IS_SYMLINK)) {
+                path = FileUtils.read_link(path);
+            }
         }
 
         FileUtils.set_contents(path, content);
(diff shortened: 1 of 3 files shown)


lib/astal/io/file.vala
```diff
@@ -85,7 +107,7 @@ public FileMonitor? monitor_file(string path, Closure callback) {
                 if (i.get_file_type() == FileType.DIRECTORY) {
                     var filepath = file.get_child(i.get_name()).get_path();
                     if (filepath != null) {
-                        var m = monitor_file(path, callback);
+                        var m = monitor_file(filepath, callback, follow_symlinks);
```
davfsa: The change from `path` to `filepath` is intentional and seems to be a bug that was never fixed, where monitoring a directory would lead to a infinite recursion (at least the way I understand it).

I can fix it as part of a different pull request if preferred

Aylur:
There should be no additional parameter for this, it should just follow them ootb.
The purpose of astal util functions is to fill the most common usage with the most sane defaults and this seems like one. I don't see anyone turning this off. If they do need to turn it off just use the glib api.

davfsa:
> There should be no additional parameter for this, it should just follow them ootb.
> The purpose of astal util functions is to fill the most common usage with the most sane defaults and this seems like one. I don't see anyone turning this off. If they do need to turn it off just use the glib api.

Noted! Will update to remove the parameter

davfsa:
@Aylur Updated!

davfsa:
This doesnt seem to work fully, investigating...

Will mark as draft until then

davfsa:
Updated to use flags (which I initially glossed over in the docs) and it is way more consistent now :)

PR #305, opened 2025-03-02
feat(hyprland): add mirrorOf member of Monitor

Dr0med4r:
For this, it is necessary to use 'monitors all' instead of 'monitors.' 
If this is not done, the 'monitoradded' event will add an empty monitor, and the 'monitorremoved' event will fail with a null error. 
This fixes issues with mirrored displays, but now mirrored monitors are listed under 'monitors' (which they previously were, but not fully and I don't know if this is intended behaviour).

### DIFF
--- a/lib/hyprland/monitor.vala
+++ b/lib/hyprland/monitor.vala
@@ -25,7 +25,7 @@ public class AstalHyprland.Monitor : Object {
     public bool vrr { get; private set; }
     public bool actively_tearing { get; private set; }
     public bool disabled { get; private set; }
-    public string mirror_of {get; private set; }
+    public Monitor? mirror_of {get; private set; }
     public string current_format { get; private set; }
     public Array<string> available_modes { get; private set; }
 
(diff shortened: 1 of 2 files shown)


lib/hyprland/monitor.vala
```diff
@@ -25,6 +25,7 @@ public class AstalHyprland.Monitor : Object {
     public bool vrr { get; private set; }
     public bool actively_tearing { get; private set; }
     public bool disabled { get; private set; }
+    public string mirror_of {get; private set; }
```
Aylur: This should be a `Monitor` type

lib/hyprland/monitor.vala
```diff
@@ -49,6 +50,7 @@ public class AstalHyprland.Monitor : Object {
         vrr = obj.get_boolean_member("vrr");
         actively_tearing = obj.get_boolean_member("activelyTearing");
         disabled = obj.get_boolean_member("disabled");
+        mirror_of = obj.get_string_member("mirrorOf");
```
Aylur: should handle if it is `null` since it was added in `0.45` this breaks backwards compatibility

Dr0med4r:
Should there be a distinction between "hyprland before 0.45" and "not mirrored"? 
If so how?

PR #323, opened 2025-03-16
AstalTray: make dbus calls async

### LINKED ISSUE #320: KeepassXC system-tray entry freezes the bar for half a minute.
**Describe the bug**
Once you've installed KeepassXC and enabled it to show in the system tray, it will hang the bar for around 30 seconds. This is not an issue if the system-tray icon setting in keepassxc is disabled.

**To Reproduce**
1. Install KeepassXC
2. Create a database (Database > New Database)
3. Enable systray icon via Settings (Gear Icon) > General > 'Show a system tray icon'
4. Restart KeepassXC
5. Input your newly created Database's password
6. Notice that the bar freezes for around 30 seconds.
 
**Expected behavior**
Bar does not freeze when the database in KeepassXC is accessed.

**Additional context**

kotontrion:
GLibs DbusProxy does make synchronous dbus calls to get properties. This can lead to freezes in some cases. This PR moves the property accessors away from DBusProxy to able to make those calls async.

closes #320

### DIFF
--- a/lib/tray/trayItem.vala
+++ b/lib/tray/trayItem.vala
@@ -147,7 +147,7 @@ internal interface IItem : DBusProxy {
 
 public class TrayItem : Object {
     private IItem proxy;
-    private List<ulong> connection_ids;
+    private bool needs_update = false;
 
 
     /** The Title of the TrayItem */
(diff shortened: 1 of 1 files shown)


Jas-SinghFSU:
Confirmed no more freezing with this PR. :tada:

PR #331, opened 2025-03-26
added brightness lib

brainlessbitch:
simple brightness lib, idk, this is needed imo


kotontrion:
here are my 2 cents:
make the backlight_model property construct only
add a BrightnessManager class, which creates creates an instance of the Brightness class for each available device.
Also leds follow the same directory structure (at /sys/class/leds/), so with this approach we can also support things like numlock, capslock, keyboard backlight, etc. See #336

Then to make usage easier, the brightness manager might also have a brightness property, which either acts as a proxy for just one  display backlight, or maybe better set the brightness for all display backlights, similar to the volume prop in the AstalWp nodes with multiple channels.

ItsLiyua:
I'd be interested in something like this

brainlessbitch:
> here are my 2 cents: make the backlight_model property construct only add a BrightnessManager class, which creates creates an instance of the Brightness class for each available device. Also leds follow the same directory structure (at /sys/class/leds/), so with this approach we can also support things like numlock, capslock, keyboard backlight, etc. See #336
> 
> Then to make usage easier, the brightness manager might also have a brightness property, which either acts as a proxy for just one display backlight, or maybe better set the brightness for all display backlights, similar to the volume prop in the AstalWp nodes with multiple channels.

i can 100% do all this, sorry for missing the comment will get the pr updated when i have time

PR #338, opened 2025-04-09
added unnamed-shell to showcase

brainlessbitch:
added [unnamed-shell](https://codeberg.org/bunbun/unnamed-shell) to showcase


brainlessbitch:
nooo, github doesn't let me cover my stupid mistakes

docs/public/showcase/unnamed-shell.webp
Aylur, line 1: I’d prefer not to put anything on the site that could be seen as controversial and a red five-pointed star is mostly associated with politics.
Would you mind changing it to something else or having the other window over the symbol?

brainlessbitch: how is it associated with politics? it's a symbol from paganism, i can do a screenshot without it, but i'd like a more in depth reason if you can give one

Aylur: Its a symbol commonly associated with communism

PR #345, opened 2025-04-23
fix(gjs/binding): check if signal is connected with `#emmiter` before disconnecting 

retrozinndev:
This checks if the binding signal is connected to `#emmiter` before disconnecting it.
It avoids warnings like this:
```
GLib-GObject-CRITICAL **: 15:37:19.666: ../glib/gobject/gsignal.c:2699: instance '0x60b2885cf340' has no handler with id '243'
```


retrozinndev:
Thanks!

PR #362, opened 2025-05-27
fix: stop_poll and arg assumption

PandeCode:
Fixed bug with variable:stop_poll
Assumption that arg is always defined, custom interpreters and runtimes like debuggers may alter this.


Aylur:
FYI the lua lib will be migrated out of this repo and won't be supported anymore.
See https://github.com/Aylur/ags/pull/685

Aylur:
You might want to open this issue over at https://github.com/tokyob0t/astal-lua

PR #407, opened 2025-10-06
Notifd update

### LINKED ISSUE #26: How to send notifications
Similar to [`Utils.notify`](https://github.com/Aylur/ags/blob/cd543afec4d9ed708990e5ba52184716809bf248/src/utils/notify.ts)

Aylur:
This update should avoid crashes and lockups when working with large number of notifications.

# Breaking changes

- `Notification.get_str_hint()` and `Notification.get_bool_hint()` are removed: use `Notification.get_hint().get_string()` and `Notification.get_hint().get_boolean()` respectively

# Changes

- use gschema instead of json files
- debounce caching
- drop json-glib and use gvariants between the daemon and proxy
- change Action struct to a class
- `send_notification` function
- `dont_ignore` and `ignore_timeout` are not persistent 

# New Features

It is now possible to send notifications using this library.

Vala example:

```vala
public async void fn() {
    var n = new Notification() {
        app_name = "My Awesome app",
        summary = "Hello There!",
        app_icon = "dialog-information-symbolic",
        image = "/path/to/an/image.png",
    };

    n.add_action(new Action("id1", "Click Me"));
    n.set_hint("custom-hint", new Variant.string("custom hint value"));

    n.invoked.connect((action_id) => {
            if (action_id == "id1") {
                print("action has been invoked\n");
            }
            n.dismiss();
        });

    try {
        yield send_notification(n);
    } catch (Error error) {
        critical(error.message);
    }
}
```

close #26


linde12:
@Aylur am i correct in that DontDisturb was removed in a7bf244e8341ae7779442d45cfdd8bc1944a6039?

I'm using Hyprpanel and had written a small application to disable notifications while screensharing. I saw that Hyprpanel had added a `set_dont_disturb` method to org.freedekstop.Notifications which i was using.

If i read correctly, this is now removed - is there any other way i can set/unset the DND? I tried `astal-notifd -t` (which is not exactly what i'm looking for, i want to set true/false not just toggle) but it did not do anything (v0.1.0, maybe bug?)

I would love to be able to set it through some dbus interface.

Aylur:
Settings have moved to dconf and as a result the dbus method has been also removed (it only existed because it was a hack to sync up notifd instances).
You can still still toggle it in code with https://aylur.github.io/libastal/notifd/property.Notifd.dont-disturb.html.
You can set it using dconf from cli `dconf write /io/astal/notifd/dont-disturb true`

PR #417, opened 2025-10-31
fix(nix): 'wrapGAppsHook' has been renamed to/replaced by 'wrapGAppsHook3'

knoopx:


ExtraBytez:
I should really read the open PRs before trying to do it myself i was literally writing this thanks for doing it!

knoopx:
btw ags flake lock file needs to be updated again so it picks the new astal with the fixes https://github.com/Aylur/ags/pull/782

PR #420, opened 2025-11-05
feat(network): add access-point-added/-removed signals to wifi

kotontrion:
This PR adds `access-point-added` and `access-point-removed` signals to the AstalNetwork.Wifi object, which allows adding/removing eg items from a Gtk.ListBox without removing and readding all of them or manually diffing the list, which would be needed in languages other than js with gnim.

Also makes the `AstalNetwork.AccessPoint.ap` property public to give the user more possibilities, eg for creating new connections, which aren't a simple password. Otherwise you would need to use libnm directly to get the Nm.AccessPoint and the match that to the AstalNetwork.AccessPoint.

### DIFF
--- a/lib/network/src/accesspoint.vala
+++ b/lib/network/src/accesspoint.vala
@@ -1,6 +1,6 @@
 public class AstalNetwork.AccessPoint : Object {
     private Wifi wifi;
-    private NM.AccessPoint ap;
+    public NM.AccessPoint ap;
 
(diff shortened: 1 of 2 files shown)


ARKye03:
Please merge this, that way I can finally drop my libnm wrapper😭

Aylur:
We could also have the wifi and wired classes implement ListModel

PR #423, opened 2025-11-14
fix(mpris): skip caching base64 cover arts

navidmafi:
This fixes thousands of lines of base64 output splattering all over your terminal. 
image

### DIFF
--- a/lib/mpris/src/player.vala
+++ b/lib/mpris/src/player.vala
@@ -680,6 +680,10 @@ public class AstalMpris.Player : Object {
             return;
         }
 
+        if (art_url.has_prefix("data:")) {
+            return;
+        }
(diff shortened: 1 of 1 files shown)


Aylur:
we should instead handle it and also write that to a file

navidmafi:
I thought to myself that since the data is available at demand at no cost (unlike https) there would be no need for caching 

Is my reasoning incorrect? I will try to adapt this to handle them

Aylur:
Its not cached for the sake of being cached. It is cached so that it is written to a file which is easier to work with on the user's side through the `cover_art` prop.

navidmafi:
whoops! sorry for forgetting about this.

PR #432, opened 2026-01-25
Deprecate some properties in favor of methods

andeston:
# Summary
- Deprecate the `focused-*` properties and replace their generated getters with explicitly declared methods that return nullable types.

# Why
- GObject properties cannot express nullability; the generated getters did not indicate they may return `null`.
- Vala properties could be nullable, but that still wouldn’t be reflected in GObject property metadata.
- Explicit getters allow us to expose correct nullable return types in the API and in generated bindings.

# What changed
- Added explicit getter methods for the focused client, workspace, and monitor.
- Those getters now express nullability.
- Marked the corresponding `focused-*` properties as deprecated and forward them to the new accessors.

# Impact
- Runtime behavior is unchanged.
- Existing code using the `focused-*` properties continues to work (properties are deprecated but still forwarded).
- Bindings will now see accurate nullability on the accessor methods, improving type correctness for language bindings.

# Considerations
- I feel that it may be appropriate to apply this property -> method(s) pattern to other properties even if there are no nullability concerns.

## Example

```vala
    public Position cursor_position {
        owned get {
            return new Position.cursorpos(message("cursorpos"));
        }
    }
```

Above does not need to be a property. I think this would make more sense.

```vala
    public Position get_cursor_position() {
        return new Position.cursorpos(message("cursorpos"));
    }
```

### DIFF
--- a/lib/hyprland/src/hyprland.vala
+++ b/lib/hyprland/src/hyprland.vala
@@ -46,10 +46,10 @@ public class Hyprland : Object {
     public List<weak Workspace> workspaces { owned get { return _workspaces.get_values(); } }
     public List<weak Client> clients { owned get { return _clients.get_values(); } }
 
-    public Monitor get_monitor(int id) {
+    public Monitor? get_monitor(int id) {
         return _monitors.get(id);
     }
-    public Workspace get_workspace(int id) {
+    public Workspace? get_workspace(int id) {
         return _workspaces.get(id);
     }
     public Client? get_client(string address) {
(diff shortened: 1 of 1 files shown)


Aylur:
They are properties so that they can be bound using GObject mechanisms. I also heavily rely on this in [ags](https://github.com/aylur/ags). Using methods+signals can technically work, but it would require a lot more boilerplate from user code in both Vala and introspecting languages. It would also limit their usage from Gtk templates and Blueprint.

It is unfortunate that properties cannot be annotated as nullable, yes, however the getter and setter methods do have nullable annotations.

So no we are not deprecating properties in favor of methods+signals. 


But these two lines are welcome changes

https://github.com/Aylur/astal/blob/2d8d441c734ffdb1bea341e73047acbcd4d6bb04/lib/hyprland/src/hyprland.vala#L49

https://github.com/Aylur/astal/blob/2d8d441c734ffdb1bea341e73047acbcd4d6bb04/lib/hyprland/src/hyprland.vala#L52

If you reverse the other changes, or open another pr I'll merge that

PR #438, opened 2026-03-09
new lib: Hyprland focus grab

Mabi19:
Adds support for the hyprland-focus-grab-v1 Wayland protocol by using astal-wl (so depends on #371).
This essentially allows making an arbitrary combination of windows to behave like a popover: clicking outside them will trigger the `clicked-away` signal on the grab context object, which would normally be used to hide the popup-like windows.
~~Only supports GTK 4.~~ Supports both GTK 3 and 4, import either version 3.0 or 4.0 respectively. This is a very simple API; it could be possible to make something more automatic by relying on window map/unmap signals but this seems like a lot of extra complexity for very little benefit.

<details>
<summary>Usage example (using ags v3)</summary>

```tsx
import { createBinding, createEffect, createState, onCleanup } from "ags"
import { Gtk } from "ags/gtk4"
import app from "ags/gtk4/app"
import AstalHyprlandFocusGrab from "gi://AstalHyprlandFocusGrab?version=4.0"

export default function Main() {
  let win: Gtk.Window
  const grabContext = new AstalHyprlandFocusGrab.GrabContext()
  grabContext.connect("cleared", () => print("click away"))

  createEffect(() => {
    grabContext.add(win)
  })
  onCleanup(() => {
    win.destroy()
  })

  const [count, setCount] = createState(0)

  return (
    <window visible name="test" application={app} $={(self) => (win = self)}>
      <box orientation={Gtk.Orientation.VERTICAL} class="main-box" spacing={4}>
        <menubutton direction={Gtk.ArrowType.RIGHT} halign={Gtk.Align.END}>
          <popover>
            <box orientation={Gtk.Orientation.VERTICAL} spacing={4}>
              <label label="popover label" />
              <button>
                <label label="popover button" />
              </button>
            </box>
          </popover>
        </menubutton>
        <label
          label={createBinding(grabContext, "active").as((v) => `active: ${v}`)}
        />
        <button
          onClicked={() => {
            grabContext.active = true
          }}
        >
          Test button
        </button>
        <button onClicked={() => setCount(count() + 1)}>
          <label label={count((c) => `Count is ${c}`)} />
        </button>
      </box>
    </window>
  )
}
```
</details>

## TODO
- [x] Nix: I don't use Nix and I don't know the syntax, so I just didn't include it for now.
- [x] Docs: Currently I put a stub. A good minimal example needs to be figured out, as well as some copy to explain what a focus grab even is. Also, this depends on another astal library, and I didn't see anything on how to write that down in the installation section. Additionally, I'd want to read through the autogenerated reference documentation and add any missing doc comments etc.
- [x] More testing: I plan to integrate this into my own setup and test that way, but that'll take a little while. I have a cold as I'm writing this, and I'm a bit busy in general.


lib/hyprland-focus-grab/src/context.vala
```diff
@@ -0,0 +1,98 @@
+namespace AstalHyprlandFocusGrab {
+internal struct WindowSignalHandlers {
+    ulong realize;
+    ulong destroy;
+}
+
+public class GrabContext : Object {
```
kotontrion, line 13: Implementing the Gtk.Buildable interace here would allow something like this in ags:

```jsx
<AstalHyprlandFocusGrab.GrabContext>
    <window .../>
    <window .../>
</AstalHyprlandFocusGrab.GrabContext>
```

lib/hyprland-focus-grab/src/context.vala
```diff
@@ -0,0 +1,98 @@
...
+                grab.commit();
+                grab.cleared.connect(handle_cleared);
+            } else {
+                if (grab == null) {
+                    return;
+                }
+                // destroy the grab
+                grab.destroy();
+                grab = null;
+            }
```
kotontrion, line 47: setting grab to null should implicitly call the destroy function, so this else can be simplified to just being `grab = null;`

lib/hyprland-focus-grab/src/context.vala
```diff
@@ -0,0 +1,98 @@
...
+        window.disconnect(handlers.destroy);
+
+        var surf = window.get_surface();
+        if ((grab != null) && (surf != null)) {
+            grab.remove(surf);
+            grab.commit();
+        }
+    }
+
+    public signal void click_away();
```
kotontrion: I would prefer calling it cleared, to better match the underlying protocol.

lib/hyprland-focus-grab/src/grab.vala
```diff
@@ -0,0 +1,74 @@
+namespace AstalHyprlandFocusGrab {
+public class Grab : Object {
```
kotontrion: What is point of separating Grab from GrabContext?

kotontrion (commented):
I did not test this myself nor did I think about this much, these are just my first thoughts after having a quick look.

Also the astal-wl code should not be included in this PR.

lib/hyprland-focus-grab/src/context.vala
Mabi19: I talked with Aylur about this and he said this style of API would be better.

kotontrion: I would like to hear about his reasoning.

Implementing the Buildable interface would not change the current API, it would just additionally allow creating the Grab like I've shown in the snippet above, which seems more in line with the ags/gnim style than creating the grab and adding the windows imperatively.

lib/hyprland-focus-grab/src/grab.vala
Mabi19: I wanted an auxiliary object to manage the underlying Wayland grab objects, and have the higher-level wrapper built on that. The Grab could be marked as internal; I left it in case someone wanted to have deeper control.

Mabi19:
> Also the astal-wl code should not be included in this PR.

I've never made a PR based on another PR, but that makes sense. I'll fix it once I get home (edit: done)

lib/hyprland-focus-grab/src/grab.vala
kotontrion: In which case it should work with wl_surfaces instead of gdk surfaces, so it could be reused in a gtk3 version of this?

Mabi19: I thought wl_surfaces weren't accessible from gjs? Maybe I'm wrong though. And that wouldn't be an issue if this was internal.

About a potential gtk3 port: this would make it much easier, yes. I don't think this could feasibly be done with one .so, you'd need hyprland-focus-grab-gtk3 and hyprland-focus-grab-gtk4. But these could probably share most of their source code, just with a small utility to get a `Gtk.Window`'s wl_surface that would be swapped out. This sounds interesting, I'll try it later

lib/hyprland-focus-grab/src/context.vala
Mabi19: obraz


obraz

Overall it's the least likely to get in the way.

Mabi19:
So I did some research on supporting GTK 3. It would theoretically be possible, except Vala doesn't ship a VAPI for GDK 3's Wayland backend :upside_down_face: 
So this library would need to maintain one of its own if we wanted to have a GTK 3 version.
But I feel that at this point only supporting 4 would be the better option, I doubt that the overlap between people that would want to use this and people who still use GTK 3 is very large.

Mabi19:
I realized I can just implement that in C. So I did! But naming conventions are weird, so I wanted to explicitly mention what this approach generates:
- `include/astal-hyprland-focus-grab-{3,4}.h`
- `lib/girepository-1.0/AstalHyprlandFocusGrab-{3,4}.0.typelib`
- `lib/pkgconfig/astal-hyprland-focus-grab-{3,4}-0.1.pc`
- `lib/libastal-hyprland-focus-grab-{3,4}.so`
- `share/gir-1.0/AstalHyprlandFocusGrab-{3,4}.0.gir`
- `share/vala/vapi/astal-hyprland-focus-grab-{3,4}-0.1.vapi`

I tried to get something close to how Astal3/4 works, but if you want a different naming convention for the different versions, that can definitely be changed.

Mabi19:
There's now a build option named `gtk-version` (type array, choices '3' and '4') that selects which GTK versions to build against.

kotontrion:
I added the nix build files. In order to make it work I had to make two changes to the mkAstalPkg function.
- pass meson compile flags 
- pass api-version

I'm still not very familiar with nix, as I rarely use it, so if there is a better way please let me know, but I think those are reasonable and straight forward changes.

Mabi19:
I tested it in my shell, and it works! It survives connecting and disconnecting monitors (and thus adding and removing windows) while a grab is active just fine as well.

lib/hyprland-focus-grab/src/manager.vala
```diff
@@ -0,0 +1,36 @@
+namespace AstalHyprlandFocusGrab {
+public bool is_supported() {
+    return !AstalWl.Registry.get_default().find_globals("hyprland_focus_grab_manager_v1").is_empty();
```
kotontrion, line 3: instead of hardcoding the interface name you could also get it from the interface struct:
```vala
return !AstalWl.Registry.get_default().find_globals(((Wl.Interface)&HyprlandFocusGrabManagerV1.iface).name).is_empty();
```
I didn't do it in the astal-wl and astal-river lib, because i like to avoid such pointer hacks, I'm just mentioning this here for completeness sake.

PR #446, opened 2026-05-12
fix(hyprland): use new lua dispatchers

retrozinndev:
As of Hyprland 0.55, [lua](https://lua.org) is the default language for configuration. Because of this change, the old dispatchers are now broken in favor of `lua` dispatchers.

In short, I changed some methods to use the new dispatchers, fixing some of the broken features(e.g.: `Workspace`, `Monitor`, `Client` focus methods), while keeping backwards compatibility with `hyprlang`.

- The `Hyprland.dispatch` method adds the `hl.dsp.` dispatcher prefix (if you think this is unnecessary, let me know!).
- Added `Hyprland.dispatch_argv` for multi-argument dispatchers if needed, both `dispatch` and `dispatch_argv` are documented, by the way.
- Added `Hyprland.eval` method for evaluating inline lua through IPC.
- Added `Hyprland:config-provider`(type `Hyprland.ConfigProvider`), which specifies whether the session is using `LUA` or `HYPRLANG` for configuration. (info is grabbed from the `status` socket command)

Also, should I add a `Client.close` method? Hyprland has two dispatchers, both for killing(force-closing) and closing a client. Currently, the `Client.kill` method calls the `window.close` dispatcher, which doesn't really match with the method's name.
If we do this, we might have to deprecate `Client.kill`(for compatibility reasons) and add a `Client.force_close()` method to actually kill the client instead of closing.

Any suggestions are welcome :D

### DIFF
--- a/lib/hyprland/src/client.vala
+++ b/lib/hyprland/src/client.vala
@@ -56,20 +56,20 @@ public class Client : Object {
     }
 
     public void kill() {
-        Hyprland.get_default().dispatch("closewindow", @"address:0x$address");
+        Hyprland.get_default().dispatch("window.close", @"address:0x$address");
     }
 
     public void focus() {
-        Hyprland.get_default().dispatch("focuswindow", @"address:0x$address");
+        Hyprland.get_default().dispatch("focus", @"{window=\"address:0x$address\"}");
     }
 
     public void move_to(Workspace ws) {
         var id = ws.id;
-        Hyprland.get_default().dispatch("movetoworkspacesilent", @"$id,address:0x$address");
+        Hyprland.get_default().dispatch("window.move", @"{window=\"address:0x$address\",workspace=$id,follow=false}");
     }
 
     public void toggle_floating() {
-        Hyprland.get_default().dispatch("togglefloating", @"address:0x$address");
+        Hyprland.get_default().dispatch("window.float", @"{window=\"address:0x$address\"}");
     }
 }
 
--- a/lib/hyprland/src/client.vala
+++ b/lib/hyprland/src/client.vala
@@ -56,20 +56,40 @@ public class Client : Object {
     }
 
     public void kill() {
-        Hyprland.get_default().dispatch("window.close", @"\"address:0x$address\"");
+        if(Hyprland.get_default().config_provider == ConfigProvider.LUA) {
+            Hyprland.get_default().dispatch("window.close", @"\"address:0x$address\"");
+            return;
+        }
+
+        Hyprland.get_default().dispatch("closewindow", @"address:0x$address");
     }
 
     public void focus() {
-        Hyprland.get_default().dispatch("focus", @"{window=\"address:0x$address\"}");
+        if(Hyprland.get_default().config_provider == ConfigProvider.LUA) {
+            Hyprland.get_default().dispatch("focus", @"{window=\"address:0x$address\"}");
+            return;
+        }
+
+        Hyprland.get_default().dispatch("focuswindow", @"address:0x$address");
     }
 
     public void move_to(Workspace ws) {
         var id = ws.id.to_string();
-        Hyprland.get_default().dispatch("window.move", @"{window=\"address:0x$address\",workspace=$id,follow=false}");
+        if(Hyprland.get_default().config_provider == ConfigProvider.LUA) {
+            Hyprland.get_default().dispatch("window.move", @"{window=\"address:0x$address\",workspace=$id,follow=false}");
+            return;
+        }
+
+        Hyprland.get_default().dispatch("movetoworkspacesilent", @"$id,address:0x$address"); 
     }
 
     public void toggle_floating() {
-        Hyprland.get_default().dispatch("window.float", @"{window=\"address:0x$address\"}");
+        if(Hyprland.get_default().config_provider == ConfigProvider.LUA) {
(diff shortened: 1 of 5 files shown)


ScarsTRF:
Out of curiosity, wouldn't it be best to make a version check for 0.55 or above of Hyprland to get those extra use cases of people who are locking their hyprland versions? Very niece I know, but still a thought for users.

I know a handful of people who are refusing to update because they rely on the normal dispatch operations and don't want to rewrite their stuff right now.

Lua might be the here on out thing for Hyprland and only have deprecated hyprlang for their config stuff, but the issue I see is with that no matter if you're using hyprlang or lua for your config on version 0.55, you're forced to rewrite ALL your hyprctl dispatch operations to lua dispatchers.

retrozinndev:
> Out of curiosity, wouldn't it be best to make a version check for 0.55 or above of Hyprland to get those extra use cases of people who are locking their hyprland versions? Very niece I know, but still a thought for users.
> 
> I know a handful of people who are refusing to update because they rely on the normal dispatch operations and don't want to rewrite their stuff right now.
> 
> Lua might be the here on out thing for Hyprland and only have deprecated hyprlang for their config stuff, but the issue I see is with that no matter if you're using hyprlang or lua for your config on version 0.55, you're forced to rewrite ALL your hyprctl dispatch operations to lua dispatchers.

Good point. Do you know any way to detect if Hyprland is using lua or not? We would need something like that(btw, detecting the hyprland config file extension is out of the question[^1]).

[^1]: It is possible to change the configuration file during Hyprland initialization with the `--config` argument, and I don't think there's a way to know where it is.

ScarsTRF:
As mentioned, this only is an issue on 0.55.0 and above, though the ipc has no way of getting hyprland version so it would have to fall back to executing `hyprctl version` on its own. You can add the -j flag to get the version info in json before grabbing the `version` value from it and storing it for checks in the module.

I would say run that `hyprctl version -j` within the get_default method AFTER checking if the env exists so you know you're not running it if it's not even existing. Get the value and just store it like so to compare when running functions.

retrozinndev:
Oh, I just found out that there's a `status` command in the hyprland socket/hyprctl that shows the config provider ;)

```zsh
$ hyprctl status -j

{
    "configProvider": "lua",
    "backend": "drm"
}
```

(I'm already working on something)

retrozinndev:
Backwards compatibility with `hyprlang` should be good now.
Could someone that is still using hyprlang please test if backwards compatibility works well?

ScarsTRF:
I can try and test in a bit on a Nix Flake and locking my version back and forth testing. I just need a little bit as I've had a lot going on yesterday and a lot going on today

retrozinndev:
> I can try and test in a bit on a Nix Flake and locking my version back and forth testing. I just need a little bit as I've had a lot going on yesterday and a lot going on today

Thanks!! No problem at all, take your time :)
Hope everything goes well for you!

navidmafi:
I tested this patch with both hyprlang and lua configs and everything worked as far as the api surface I'm using (can be found below)

<details>

```ts
import { createBinding, For } from "ags/index";
import { Gtk } from "ags/gtk4";
import AstalHyprland from "gi://AstalHyprland?version=0.1";

export default () => {
  const hypr = AstalHyprland.get_default();
  const workspaces = createBinding(hypr, "workspaces");
  const focused = createBinding(hypr, "focusedWorkspace");
  return (
    <Gtk.Box valign={Gtk.Align.CENTER}>
      <Gtk.EventControllerScroll
        onScroll={(_c, _dx, dy) => {
          const spaces = [...hypr.workspaces].sort((a, b) => a.id - b.id);
          const len = spaces.length;
          if (len === 0) return;

          if (!focused) return;
          const curIdx = spaces.findIndex((ws) => ws.id === focused.get().id);
          if (curIdx === -1) return;

          // convert dy to an integer step (1 or -1)
          const step = Math.sign(dy) || 0;
          if (step === 0) return;

          const nextIdx = (curIdx - step + len) % len; // -dy in original; here dy sign already accounts
          const target = spaces[nextIdx];
          if (!target) return;

          // avoid redundant .focus() calls
          if (target.id === focused.get().id) return;
          target.focus();
        }}
        flags={Gtk.EventControllerScrollFlags.VERTICAL}
      />
      <Gtk.Box spacing={2}>
        <For
          each={workspaces.as<Array<AstalHyprland.Workspace>>((wss) =>
            wss.filter((a) => a.id > 0).sort((a, b) => a.id - b.id)
          )}
        >
          {(ws) => (
            <Gtk.Button
              class={focused.as((fw) =>
                [
                  "p-0 min-w-8 min-h-8",
                  fw?.id === ws.id ? "bg-primary" : "bg-surface",
                ].join(" ")
              )}
              label={ws.id.toString()}
              onClicked={() => ws.focus()}
            />
          )}
        </For>
      </Gtk.Box>
    </Gtk.Box>
  );
};
```

</details>

ScarsTRF:
Oh apologies, I though I sent my information about in a while back. Either I didn't hit comment or something, but yes it did work for me when I was testing on my shell config.

retrozinndev:
> Oh apologies, I though I sent my information about in a while back. Either I didn't hit comment or something, but yes it did work for me when I was testing on my shell config.

That's okay! Glad it works as expected ;D

PR #456, opened 2026-06-28
fix(bluetooth): signals when removing bluetooth devices and adapters

### LINKED ISSUE #455: AstalBluetooth's devices and adapters arrays emit signal before deleting their elements
**Describe the bug**
AstalBluetooth's devices and adapters arrays are updated after the signals `device-removed` and `adapter-removed` respectively are emitted which causes anything to be dependent on their values to be updated whenever the next signal is emitted.

**To Reproduce**
```tsx
const bluetooth = AstalBluetooth.get_default();
bluetooth.connect("device-removed", (self, dev) => {
    console.log("deleted device: ", dev.name);
    console.log("without timeout: ", bluetooth.devices.map((d) => d.name).join(", "));
    setTimeout(() => console.log("with timeout: ", bluetooth.devices.map((d) => d.name).join(", ")), 1000);
});
```
The deleted device is present when console logged without timeout, but is deleted when logged without the timeout meaning that it was deleted sometime after the signal.
The issue is similar with adapters.

**Expected behavior**
Deleted device or adapter should be removed from the array before the signal is emitted.

DistortedDragon1o4:
Fixes #455

### DIFF
--- a/lib/bluetooth/src/bluetooth.vala
+++ b/lib/bluetooth/src/bluetooth.vala
@@ -184,14 +184,16 @@ public class AstalBluetooth.Bluetooth : Object {
     private void on_interface_removed (DBusObject object, DBusInterface iface) {
         if (iface is IDevice) {
             unowned var device = (IDevice)iface;
-            device_removed(_devices.get(device.g_object_path));
+            unowned var device_obj = _devices.get(device.g_object_path);
             _devices.remove(device.g_object_path);
+            device_removed(device_obj);
         }
 
         if (iface is IAdapter) {
             unowned var adapter = (IAdapter)iface;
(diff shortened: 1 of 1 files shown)


Aylur:
I'm a bit confused about ownership here. Who owns the device after its removed from the hashmap? Wouldn an unowned reference potentially mean that a destroyed object is passed to the signal?

DistortedDragon1o4:
Would it be better to transfer the ownership of the object to a private variable and then pass that on?

Aylur:
Just removing the `unowned` keyword should do the trick. Why did you make it `unowned`?

DistortedDragon1o4:
Oh it's because I have 0 idea of how vala works

PR #459, opened 2026-07-14
new lib: workspace

Mabi19:
A binding for the ext-workspace-v1 Wayland protocol.

This protocol exposes workspaces, as well as groups, which are annoyingly abstract collections of workspaces (belonging to an arbitrary number of outputs/monitors). The library exposes 1:1 bindings for these as GObjects, as well as helper MonitorView objects for the common case of listing all the workspaces on a given output/monitor. It also includes a full pending state mechanism, which is handled properly (new state is withheld until the compositor sends `done`, and `commit` must be manually called by the user when calling the relevant workspace/group methods, which I will call "action methods" from now on)

The library depends on GTK 4. Support for GTK 3 would not be too hard to implement but would be annoying.

Disclaimer: I used Kimi K2.7 Code while making the library, mostly to check my work for copy/paste errors etc. All of the code is written by hand.

## TODO
- [x] CLI: Currently there's some basic code for listing workspaces and groups. ~~This needs to be expanded.~~ There is now a full CLI, and I tested as much of it as I could (not all of the exposed functionality is supported by my compositor)
- [x] Docs: I wrote the docs page, although the API is theoretically not final I don't think it's gonna change THAT much.
- [x] Nix: I don't speak Nix, so I did not include it. But this should be fairly easy to Nix up.
- [x] An API choice to consider: should there be a mechanism to automatically commit after action methods are called? Manual control might still be desired, so I'd pair this with a `freeze_autocommit`/`thaw_autocommit` for when you want to do several changes atomically. (I ended up adding this as described)
- [x] Testing: I rewrote my own workspace indicators to use the library, which uncovered the below two bugs.
- [ ] There's an issue where calling workspace action methods in a GTK setting causes some event loop weirdness that makes the next input event delayed. ~~I have no idea why it happens.~~ I believe it's [caused by GTK](https://gitlab.gnome.org/GNOME/gtk/-/work_items/8331).
- [x] There's an issue where the `for_monitor` function fails to find the monitor occasionally. It needed to be fixed in AstalWl.


Mabi19:
I'm currently moving my own workspace indicator to AstalWorkspace, and I've hit a snag with AstalWl. Namely, even with the GDK integration, I can't just get `AstalWl.Output`s from the `wl_output`s that `Gdk.Monitor`s have. This is because they are from different registries, since GDK's registry is not accessible; @kotontrion how should I get an AstalWl.Output from a GDK monitor?

kotontrion:
In Gdk4 you can get the monitors connector name, which is the same as the astal.output name, so you can use the find_output_by_name function.

I thought about getting the wl_outputs from gdk, but this would require splitting the lib into separate gdk3 and 4 versions, which I want to avoid, also getting the wl_output from gdk is not even possible in gdk3.

I haven't looked at the code, but why do you depend on gtk4? Any way to make it gtk/gdk independent?

Also you rewrote the git history, this PR includes rewritten commits already in main. Please don't._

Mabi19:
> I haven't looked at the code, but why do you depend on gtk4? Any way to make it gtk/gdk independent?

There's a utility function to filter the workspaces by monitor. This is such a common case that I don't think it's worth not depending on GTK4 for. Essentially, pretty much everyone would need to include that same function in their codebase. And Aylur says depending on GTK4 is fine.

> Also you rewrote the git history, this PR includes rewritten commits already in main. Please don't.

I don't know why the github sync fork button did that. I'll try to figure out how to not. \
EDIT: I figured out how to not.

Mabi19:
I fixed the monitor-to-view function, and I now have workspaces! With a lot less client code than before!
2026-07-18-212514-spaceshot

I also added autocommitting, with a mechanism similar to the emission of notify signals: `freeze_autocommit()` and `thaw_autocommit()` which work analogously to notify signal emission.

Unfortunately, there's a gnarly issue lodged somewhere deep in the event loop: calling `workspace.activate()` (which boils down to sending an `activate` on it and `commit` on the manager) from within an event controller handler (like by clicking my workspace buttons, or scrolling on them) breaks GTK's event delivery: the next mouse event arrives delayed, with the mouse event after it (so, in my case, scrolling after clicking a workspace would have the scroll delayed until the mouse is moved)

I also tried forcing AstalWl to use its own wayland connection (by replacing the call to get_wl_display() in the registry constructor with a null). This did not have any effect, which is REALLY WEIRD.

I vaguely remember experiencing something like this in the past and fixing it somehow, but I don't remember how or when. Guess I'm gonna look through my commit history now.

EDIT: I couldn't find anything. I genuinely have no clue how to approach this anymore D:
at least AstalWorkspace itself works fine?

Mabi19:
I'm experiencing occasional crashes when my monitors turn off and on due to their DPMS being wonky. Unfortunately they're really shy and stop happening when I'm trying to investigate >:(

EDIT: I figured it out :D
In the workspace destroy function, I removed all the references to the workspace, and then did things with it after it was destroyed. Therefore, kaboom.

Mabi19:
Hyprland 0.57 will remove the .conf config format, which will also break AstalHyprland as it is right now (switching to lua also makes dispatchers use lua). So, either this or #446 should be merged before then so that it's still possible to make workspace indicators on Hyprland.

Mabi19:
Eldritch bug update! I have found the code in GTK that causes the scroll delays. https://gitlab.gnome.org/GNOME/gtk/-/work_items/8331
The bug with `for_monitor` not finding the Wayland output is mine and still not fixed, although it should be much simpler to deal with than that one.

Mabi19:
The above commit is called "Fix group initialization upon new monitor creation". However, the `for_monitor` function still doesn't work for reasons which are not my doing!

# In short
(lib/wl/output.vala)
obraz

# The long version
I filter the outputs a group is added to to only the ones AstalWl knows about (since they're nicely wrapped in GObjects, so non-Vala code can safely interact with them). To achieve this I use `Registry.get_output_by_wl_output()`:
```vala
private void handle_output_enter(ExtWorkspaceGroupHandleV1 handle, Wl.Output wl_output) {
    var output = AstalWl.get_default().get_output_by_wl_output(wl_output);
    if (output == null) {
        // This is probably an output from the GDK registry.
        return;
    }
    debug("output %p enter group %p", output, handle);
    pending_added_outputs.add(output);
}
```

However, when a new output is created, this returns null. Why? Let's look at the WAYLAND_DEBUG log:
<details>
<summary>WAYLAND_DEBUG log of a new monitor being added</summary>
Filtered to only show relevant entries, and annotated (after --). I also log all group-modifying actions manually.

```
-- AstalWl's registry gets it as wl_output#359.
[3316628.903] {Default Queue} wl_registry#47.global(72, "wl_output", 4)
[3316628.911] {Default Queue}  -> wl_registry#47.bind(72, "wl_output", 4, new id [unknown]#359)
-- GDK's registry gets it as wl_output#353.
[3316628.915] {Default Queue} wl_registry#2.global(72, "wl_output", 4)
[3316628.918] {Default Queue}  -> wl_registry#2.bind(72, "wl_output", 4, new id [unknown]#353)
-- A new workspace group is created for this monitor, and I receive it.
[3316628.938] {Default Queue} ext_workspace_manager_v1#51.workspace_group(new id ext_workspace_group_handle_v1#4278190080)
(mabi-shell:115552): astal-workspace-DEBUG: 21:06:41.445: manager.vala:294: create group 0x55c2089a9c60
[3316628.948] {Default Queue} ext_workspace_group_handle_v1#4278190080.capabilities(0)
[3316628.950] {Default Queue} ext_workspace_group_handle_v1#4278190080.workspace_enter(ext_workspace_handle_v1#4278190085)
(mabi-shell:115552): astal-workspace-DEBUG: 21:06:41.445: group.vala:147: workspace 0x55c207f9e340 enter group 0x55c2089a9c60
[3316628.958] {Default Queue} ext_workspace_group_handle_v1#4278190080.workspace_enter(ext_workspace_handle_v1#4278190086)
(mabi-shell:115552): astal-workspace-DEBUG: 21:06:41.445: group.vala:147: workspace 0x7efc14027960 enter group 0x55c2089a9c60
[3316630.189] {Default Queue} ext_workspace_manager_v1#51.done()
(mabi-shell:115552): astal-workspace-DEBUG: 21:06:41.446: manager.vala:206: manager done
-- The AstalWl output gets added to the group.
-- This isn't recognized by my code, because at this moment the output is NOT in AstalWl's list.
-- (If it had been, there would've been an "output X enter group Y" log below this)
[3316631.441] {Default Queue} ext_workspace_group_handle_v1#4278190080.output_enter(wl_output#359)
-- The GDK output gets added to the group.
[3316657.757] {Default Queue} ext_workspace_group_handle_v1#4278190080.output_enter(wl_output#353)
[3316657.776] {Default Queue} ext_workspace_manager_v1#51.done()
(mabi-shell:115552): astal-workspace-DEBUG: 21:06:41.474: manager.vala:206: manager done
-- I set up this debug log to fire when AstalWl fires the output_added signal (and it's added to the Registry's output list)
-- Notice that my code only receives the output AFTER the compositor has done things with it.
(mabi-shell:115552): astal-workspace-DEBUG: 21:06:41.479: manager.vala:182: immediate add_named_output for 0x55c208a054c0 (DP-2)
```
</details>

That's weird. Why does the output get added to AstalWl's list so late, after the compositor has already done things with it?
At this point I looked at the AstalWl code, namely at the AstalWl.Output constructor.

```vala
internal Output(Global global, Wl.Registry registry, Wl.Display display, ZxdgOutputManagerV1 output_manager) {
    Object(id: global.name);
    this.geometry = Rectangle();
    this.output_geometry = Rectangle();
    this.output = registry.bind<Wl.Output>(global.name, ref wl_output_interface, uint.min(global.version, 4));
    this.output.add_listener(output_listener, this);
    display.roundtrip(); // Uh oh.
    if (output_manager != null) init_xdg(output_manager, display);
}
```
The roundtrip effectively makes this function async: you're ceding control back to the Wayland event loop until sometime later. But the output *has already been bound* 2 lines above, so the compositor is gonna be sending messages using it during the roundtrip, like our `[3316631.441] {Default Queue} ext_workspace_group_handle_v1#4278190080.output_enter(wl_output#359)`.
The AstalWl.Output object isn't even done constructing yet when the output_enter is sent. So, my code which handles this asks AstalWl during the roundtrip, receives nothing, and assumes the output is GDK's (or someone else that does wayland, like mesa, although I haven't experienced mesa bind outputs).

In practice, this means the group never enters the new output for library consumers. (Although my code was separately broken before the last commit, and it also used to roundtrip, so it would've broken future AstalWl-based libraries)

@kotontrion Sorry for the wall of text, but unfortunately AstalWl is implicated in this issue.

kotontrion:
Iirc I added it because of the xdg output support. The wl_outputs globals might be announced before the xdg output manager, this roundtrip makes sure the xdg output manager was registered at this point if the compositor supports it. We can remove that roundtrip in that constructor by rewriting how xdg outputs are handled. Basically when the xdg output manager gets announced, we need to iterate over all our outputs and init the xdg output for that wl output. This would have the side effect that the newly created output might have different size/position/etc properties, which will the change right after when the xdg manager is registered.

I can look at this on the weekend.

Mabi19:
To fix the `for_monitor` issue I now depend on PR #465's extended AstalWl API. Because the only bug is in GTK, I think it's fine to mark this as ready for review.

lib/workspace/src/group.vala
```diff
@@ -0,0 +1,245 @@
...
+     */
+    public Object? get_item(uint position) {
+        if (position >= workspaces.length) {
+            return null;
+        } else {
+            return workspaces[position];
+        }
+    }
+
+    public Type item_type {
```
Aylur, line 71: Isnt it only the methods required for ListModel. Or if it is, can't we only write the properties, since the methods get auto generated?

lib/workspace/src/manager.vala
```diff
@@ -0,0 +1,321 @@
...
+    private GenericArray<WorkspaceGroup> pending_created_groups;
+    private GenericArray<WorkspaceGroup> pending_deleted_groups;
+
+    private int autocommit_freeze_count;
+    private bool autocommit_pending;
+
+    /**
+     * Emitted when the compositor updates the workspace state.
+     */
+    public signal void updated();
```
Aylur, line 50: I know the protocol has an `updated` signal, but I don't think we should have one in the lib. Users can already use GObject notify signals

lib/workspace/src/manager.vala
```diff
@@ -0,0 +1,321 @@
...
+
+#endif
+
+    /**
+     * Commit any pending workspace method calls.
+     * By default, this is done automatically; you can use freeze_autocommit()
+     * to delay committing any changes until thaw_autocommit() is called.
+     * You can also call this while autocommits are frozen to commit manually.
+     */
+    public void commit() {
```
Aylur, line 106: What is a use case for freezing it? I would expect the whole lib to hide wayland communication from me and auto commit asynchronously whenever a method is called or a property is assigned

lib/workspace/src/monitorview.vala
```diff
@@ -0,0 +1,241 @@
...
+    private WorkspaceManager manager;
+    private AstalWl.Output? output;
+    /**
+     * Whether the Wayland output this object was created with still exists, and thus the object will still receive updates normally.
+     */
+    public bool valid { get; private set; default = true; }
+    /**
+     * Emitted when the Wayland output this object was created with gets destroyed.
+     */
+    public signal void invalidate();
```
Aylur, line 16: Does this need to exist? Isn't this already covered by `notify::valid`?

lib/workspace/src/manager.vala
Mabi19: Let's say you have river classic-like workspaces (arbitrarily many can be active at once), and you want to deactivate one and activate another. The protocol has commits be separate so you can do both of these at once, atomically. With AstalWorkspace, you'd:
```vala
manager.freeze_autocommit();
workspace1.deactivate();
workspace2.activate();
manager.thaw_autocommit();
```
This is a bit of a contrived example but it's useful anytime you want to do multiple things from one UI action.

lib/workspace/src/monitorview.vala
Mabi19: I followed `GdkMonitor` which also has both of these. It is technically unnecessary and can be removed if you want

lib/workspace/src/group.vala
Mabi19: The [ListModel documentation](https://docs.gtk.org/gio/iface.ListModel.html) says:
> Over time, it has established itself as good practice for list model implementations to provide properties item-type and n-items to ease working with them. While it is not required, it is recommended that implementations provide these two properties. They should return the values of [g_list_model_get_item_type()](https://docs.gtk.org/gio/method.ListModel.get_item_type.html) and [g_list_model_get_n_items()](https://docs.gtk.org/gio/method.ListModel.get_n_items.html) respectively and be defined as such:
> ```c
> properties[PROP_ITEM_TYPE] =
>   g_param_spec_gtype ("item-type", NULL, NULL, G_TYPE_OBJECT,
>                       G_PARAM_CONSTRUCT_ONLY | G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS);
> properties[PROP_N_ITEMS] =
>   g_param_spec_uint ("n-items", NULL, NULL, 0, G_MAXUINT, 0,
>                      G_PARAM_READABLE | G_PARAM_STATIC_STRINGS);
> ```
From my testing, Vala does not recognize the autogenerated methods from these as fulfilling the interface, so they have to be written manually.

lib/workspace/src/manager.vala
Mabi19: The CLI uses `updated` in watch mode. Doing it with plain `notify`s and listening to the list models would also be possible, but really annoying. Or this signal could be made internal if you don't want it exposed.

kotontrion: The issue with the notify signals, is that you either need to specify a property, which might not get updated in the done handler, so you miss updates, or you use the generig notify signal, but that gets emitted multiple times, once for each updated property, so your signal handler gets invoked multiple times and the users need to debounce it themselves.

lib/workspace/src/cli.vala
```diff
@@ -0,0 +1,435 @@
...
+                } else {
+                    builder.add_value(workspace_to_json(workspace, monitors));
+                }
+            }
+            print("%s\n", Json.to_string(builder.end_array().get_root(), pretty.enabled));
+        }
+
+        public override int execute() {
+            var manager = AstalWorkspace.get_default();
+            AstalWl.get_default().get_display().roundtrip();
```
kotontrion: use `AstalWl.get_default().roundtrip();` instead

lib/workspace/src/cli.vala
```diff
@@ -0,0 +1,435 @@
...
+                }
+                if ((w_name.value != null) && (workspace.name != w_name.value)) {
+                    continue;
+                }
+                if ((workspace.capabilities & AstalWorkspace.WorkspaceCapabilities.ACTIVATE) == 0) {
+                    found_ineligible = true;
+                    continue;
+                }
+                workspace.activate();
+                display.flush();
```
kotontrion, line 202: Why is the flush needed here and in some other execute functions?

lib/workspace/src/manager.vala
```diff
@@ -0,0 +1,321 @@
...
+     * Get a proxy object which filters workspaces to those that belong to a group on the specified Wayland output.
+     */
+    public WorkspaceMonitorView for_output(AstalWl.Output output) {
+        return new WorkspaceMonitorView(this, output);
+    }
+
+    // Having GDK linked in makes AstalWl use it (and print a critical if there is no display).
+    // Pretty much everyone will have it linked in anyway, EXCEPT the astal-workspace CLI.
+    // So add an escape hatch to stop depending on it.
+#if !NO_GTK
```
kotontrion: When gdk was linked but not initialized, AstalWl did result in a critical warning, but still fell back to its own connection. Since #465 the critical warning was circumvented. This still makes sense, so the cli does not depend on gdk, but the comment needs to be updated then.

lib/workspace/subprojects/wl-vapi-gen.wrap
```diff
@@ -0,0 +1,8 @@
+[wrap-file]
+directory = wl-vapi-gen-1.0.0
+source_url = https://github.com/kotontrion/wl-vapi-gen/archive/refs/tags/1.0.0.tar.gz
+source_filename = wl-vapi-gen-1.0.0.tar.gz
+source_hash = 28a130fa59c883f23c3aaaacd326a7c42bdc449dc809103814db9c077934000f
```
kotontrion: update to wl-vapi-gen 1.1.0

lib/workspace/src/manager.vala
Mabi19: This is gonna be externalized into AstalWl4 anyway, so I'll remove it completely once that exists.

lib/workspace/src/cli.vala
Mabi19: My thinking was that the CLI is going to exit soon, so the Wayland connection is going to close, and AstalWl isn't built for cleaning it up properly, so the flush is a safety net to make sure the commands reach the compositor and aren't stuck in some intermediate buffer.

kotontrion: The get_default functions are designed yo stay alive until the process gets terminated, so they do not do cleanup, its the same design that gdk uses. But we could add atexit handlers to flush the display and clean it up before exit. This would need to be added to all WL libs then so they can destroy their objects before the display disconnect, otherwise it will cause issues.

kotontrion: I just noticed inconsistencies, because the flush gets called in some commands but not others, hence the question, but on second look, it is only flushed in the commands that write to the wl display socket, so that's fine then.

Mabi19:
I copied the GDK-to-Wayland output matching logic into a separate library: #470. Unfortunately I don't have the time to remove the corresponding code from here at the moment, I'll do it later.

lib/workspace/src/manager.vala
Aylur: I'm generally not a fan of signals like this because reactivity should be composed of values you depend on instead of a "catch all" signal. But I guess this is fine because its annoying to setup listeners for nested objects.

lib/workspace/src/manager.vala
Aylur: Is it possible to enter into an invalid state between commits? If it is, is it handled on the server side, or do we need to handle it on our client side? For example a mutating an already "deleted" workspace?

lib/workspace/src/monitorview.vala
Aylur: Lets keep it then

lib/workspace/src/manager.vala
Mabi19: No. Deleting Wayland objects (in modern protocols after this stuff was figured out) is always done by the client (in this case when the last reference to the workspace object is dropped); when the server sends a `removed` event, all it does is make the object inert and all further usage will do nothing. It's like this to prevent race conditions, but it also works in our favor here.

In fact the ext-workspace-v1 protocol does not even define any errors - the only thing is sending requests after `stop` on the workspace manager, but that's not accessible via AstalWorkspace

PR #465, opened 2026-08-01
AstalWl: fix roundtrips

kotontrion:
as discussed in #459 some roundtrip calls in the AstalWl lib can cause issues, so i removed them, I also added an async roundtrip function, which might come in handy in some cases.

#### Changes:
- the AstalWl.Output properties are not updated atomically once the done event is received
- removed the roundtrip in the AstalWl.Output constructor, therefore when the `output-added` signal is emitted, the outputs properties are not yet set. This might break some behaviour in code that relies on that.
- added LsitModel api to the registryy for outputs and seats. Outputs are only added to the list after a `done` event is received, so after its properties have been set, mimicking the old behaviour of the `output-added` signal.
- add an `ASTAL_WL_DISPLAY` environment variable which affects the wayland connection using these values:
    - gdk or gtk: force the use of gdks wayland connection and do not fall back to creating our own
    - astal or any other value: create our own wayland connection and do not use gdks connection
    - unset (default): try to use gdks connection and fallback to creating our own if that fails
 - fix the critical warning if gdk is linked but not initialized
 - add a `roundtrip` and `roundtrip_async` function to allow doing syncs from gjs

### DIFF
--- a/lib/river/src/river.vala
+++ b/lib/river/src/river.vala
@@ -73,6 +73,7 @@ public class River : Object {
 
     private void handle_output_added(AstalWl.Output wl_output) {
         var output = new Output(wl_output, this, river_status);
+        output.output.invalidate.connect((o) => this.handle_output_removed(o));
         this._outputs.append(output);
         this.output_added(output);
         this.notify_property("outputs");
@@ -181,7 +182,7 @@ public class River : Object {
         };
         CallbackResult result = CallbackResult() {};
         cb.add_listener(cb_listener, &result);
-        this.astal_registry.get_display().roundtrip();
+        this.astal_registry.roundtrip();
         output = result.msg;
         return result.success;
     }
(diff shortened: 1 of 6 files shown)


Mabi19:
Since outputs are no longer guaranteed to have all the data immediately after appearing to user code, it'd be nice if AstalWl also exposed a signal like `output-added` but emitted once the output sends its first `done` (i.e. it's got all the data). I currently have an implementation of this in AstalWorkspace, so I can have outputs ready for matching to GDK monitors. Even if that logic ends up in a dedicated GDK-to-AstalWl matching library, I can see this being convenient overall.

kotontrion:
Yeah, that's the reason this is a PR and not fixed on main. I plan to do some more changes in this PR.
- I will make use of the done signals to make the changes of the property updates atomic. I will also introduce a dedicated signal on the output object emitted on changes.
- Add ListModel APIs. I planned to add the objects to the list after they were updated, which is basically what you're asking here.
- Add an env var to force the use of the internal WL display instead of using gdks, which could be useful for debugging.
- have another look at which other wl objects could be useful and should be bound in AstalWl

kotontrion:
@Mabi19 please check if these changes fix your issues in the workspace lib and also test if this broke anything in your HyprlandFocusGrab lib. Since neither of these libs are merged yet, updating them don't block this pr, but I would need to know if something  changed here needs to be fixed in AstalWl then. The River lib is also updated n this PR to address the different timings in the output creation.

I plan to merge this within the next week if no issues arise.

Mabi19:
HyprlandFocusGrab works fine, but ~~Workspace breaks entirely (or at least the monitor views do). I'm gonna investigate this now~~

EDIT: I fixed it! Turns out there was an edge case in the order outputs initially appeared that the roundtrip hid initially but it isn't there anymore. I still need to check if Workspace still works upon monitor shuffling (the original failure case)

Mabi19:
I confirmed that both HyprlandFocusGrab and Workspace work fine with these changes, and that Workspace's new output handling issue is also resolved.

PR #470, opened 2026-08-16
new lib: AstalWl4

Mabi19:
A helper library for matching GTK 4's GDK monitors to Wayland outputs.

This logic was originally in AstalWorkspace, but it could be useful elsewhere and it's nicer having the GTK-dependent bits outside so they could be swapped out.

The API is a single async function `public async AstalWl.Output? find_output_by_gdk_monitor(Gdk.Monitor monitor)`.

Uncrustify is being really weird and indenting the contents of `find_output_by_gdk_monitor` wrong. I have no idea why it's doing that. Also it's not adding newlines at the end of files for some reason.

Depends on #465.

This PR also moves lib/wl into a separate folder for all the basic Wayland libraries; it lives in lib/wl/wl (and this binding is in lib/wl/wl4). I also rewrote the path in flake.nix but I don't know if there's something else I'd need to do to make nix work.

Needs Nix.


lib/wl/wl4/src/match.vala
```diff
@@ -0,0 +1,105 @@
...
+
+        registry = AstalWl.get_default();
+        var output = registry.get_output_by_name(monitor.connector);
+        if (output != null) {
+            debug("immediate finish");
+            this.output = output;
+            finish();
+        } else {
+            debug("delayed finish");
+            registry_connect_add_output = registry.output_list_model.items_changed.connect(this.handle_new_output);
```
kotontrion, line 68: When for some reason the output never gets registered in AstalWl, handle_new_output will never be called and the the async function will never be resolved. Well, it will be resolved when you disconnect the monitor and the Gdk monitor gets invalidated.

I would suggest making an async roundtrip and then return null if the output still hasn't registered by then. Note that you might need two roundtrips to make sure.
I would suggest the following flow:
if output not found in astal -> async roundtrip -> now astal should know bout it, check the registries output array and either find the correct output, or an output with an unset name, in which case do another async roundtrip (or connect to the monitors changed signal), otherwise there's an issue, so resolve with null

Mabi19: I don't think there is a way for this to theoretically happen...?

I mean, I could add a general failsafe here: do a couple (I think 3 would be a good number?) of roundtrips that start during the first async wait and cancel if they finish.
In general I would like to avoid relying on roundtrips because in some previous experiments which were based on them they worked inconsistently in the failure case that prompted this library to be so complex.

Not sure if I'll be able to do this today.

kotontrion: If all involved software is implemented correctly, it shouldn't happen, but there are so many different paths, that I'd like to be defensive here.

I don't like having the signal based workflow and the roundtrip simultaneously. Can you describe the issues you ran into using roundtrips?

lib/wl/wl4/meson.build
```diff
@@ -0,0 +1,103 @@
...
+)
+
+version = meson.project_version()
+name = meson.project_name()
+version_split = version.split('.')
+api_version = '0.1'
+gir_name = 'AstalWl4-0.1.gir'
+typelib_name = 'AstalWl4-0.1.typelib'
+
+add_project_arguments('-DG_LOG_DOMAIN="astal-wl4"', language: 'c')
```
kotontrion: The only libs using this currently are AstalWl and AstalWp, both of which use PascalCase as the logdomain, while you use kebab case. I don't think there are any conventions on this, but we should be consistent within astal.

lib/wl/wl4/src/match.vala
Mabi19: Yeah, being defensive would be good here.

Unfortunately I don't remember the roundtrip issues too well because it was a while ago, but it was approximately like this:
1. Due to the compositor being confused by the monitor having weird DPMS it is quickly created and then immediately destroyed.
2. GDK sees the created monitor and makes a Gdk.Monitor for it, and hands it off to user code.
3. I see the monitor has no data and roundtrip. During the roundtrip the output object is destroyed, which caused stuff to break. I think with the newer async roundtrip stuff this MAY be fine (as in, the code can handle it)? Not sure. But intermittently I also experienced Wayland protocol errors originating from somewhere within GDK during the roundtrip, and I don't fully understand those.

So, in short, I think a fully roundtrip-based approach could work, but I don't really know. Sorry for my lack of a concrete answer.

On a more conceptual level, I don't like relying on roundtrips here because they are imprecise with consuming the event stream and are a bit too implicit for my taste.

lib/wl/wl4/meson.build
Mabi19: Renamed to AstalWl4.

Issue #473, opened 2026-08-18
add support for `ext-idle-notify-v1` in `AstalWl`


Muqtxdir:
**Is your feature request related to a problem? Please describe.**

`AstalWl` gives me outputs and seats, but it does not tell whether the user has gone idle.  Anything that wants to dim/blank/lock the screen after a while has to open its own wayland conn just for that, next to the one `AstalWl` already has.

**Describe the solution you'd like**
add idle support in `lib/wl`, through `ext-idle-notify-v1` protocol

Muqtxdir:
hey @Mabi19 given that in your pr #470 you are refactoring astal-wl and splitting appropriately, can I request you to consider implementing the idle aspect for outputs and integrate it within your pr?

Mabi19:
No. Implementing protocols is not the point of my PR at all; in fact the only thing it does to AstalWl itself is move it into a different directory.

Muqtxdir:
gotcha, yeah makes sense. (for some reason GitHub is not showing me pr's diff and made the assumption based on description that it is kinda refactoring the astal-wl lib but now that you explained the direction - my bad)

Muqtxdir:
tysm!

PR #475, opened 2026-08-23
feat (`astal-wl`): add `ext-idle-notify` support

Muqtxdir:
### description

- adds the staging `ext-idle-notify-v1` protocol 
- adds `IdleNotification` that emits `idled` and `resumed`, and tracks an `idle` property.
- integarte with `Seat` and `Registry` 

Related to issue: #473


Muqtxdir:
apologies for the noise here, spend an afternoon to figure this out only to realize there was already a PR for this: #474 that handles it differently+better

PR #480, opened 2026-08-31
fix(tray): handle combined service and object path in StatusNotifierWatcher

VikingOwl91:
### Summary

Some StatusNotifierItem implementations (most notably Chromium and Electron applications such as Element, Slack, Discord, etc.) register against `org.kde.StatusNotifierWatcher` using a combined bus name and object path string formatted like:
`org.freedesktop.StatusNotifierItem-<PID>-<ID>/StatusNotifierItem/<ID>`

Previously, `RegisterStatusNotifierItem` in `watcher.vala` only checked whether `service[0] == '/'`. If it did not start with `/`, it assumed the string was solely a bus name and unconditionally appended `/StatusNotifierItem`.

This produced malformed paths like:
`org.freedesktop.StatusNotifierItem-.../StatusNotifierItem/1/StatusNotifierItem`

Because `TrayItem` attempts to read DBus properties and menu models from that path, and Chromium only exposes `/StatusNotifierItem/1`, all properties (`id`, `title`, `gicon`, `menu_model`, `action_group`) resolved to `null`, resulting in an empty non-interactive tray icon.

### Solution

Check if `service` contains a `/` and split the bus name and object path accordingly:
- If `service` starts with `/`: sender is bus name, `service` is path
- If `service` contains `/`: split on first `/` into bus name and path
- Otherwise: `service` is bus name, default path is `/StatusNotifierItem`

### DIFF
--- a/lib/tray/src/watcher.vala
+++ b/lib/tray/src/watcher.vala
@@ -46,6 +46,10 @@ internal class StatusNotifierWatcher : Object {
         if (service[0] == '/') {
             path = service;
             busName = sender;
+        } else if (service.contains("/")) {
+            var parts = service.split("/", 2);
+            busName = parts[0];
+            path = "/" + parts[1];
         } else {
             busName = service;
             path = "/StatusNotifierItem";


Somfic:
can confirm this fix will work:

Tested with Mullvad VPN 2026.4 (Electron 43), astal-tray 0.1.0 under an AGS/GTK4 shell. 
It registered service string as `org.freedesktop.StatusNotifierItem-65100-1/StatusNotifierItem/1`, which previously fell in the `else` branch.

With this it correctly extracts `busName`

kotontrion:
all of the electron based apps i just tried worked fine. Mullvad is working for me as well. 
Please run this command, then start the affected apps:
```bash
dbus-monitor --session "type='method_call',interface='org.kde.StatusNotifierWatcher',member='RegisterStatusNotifierItem'"
```
and share its output.

VikingOwl91:
> all of the electron based apps i just tried worked fine. Mullvad is working for me as well. Please run this command, then start the affected apps:
> 
> ```shell
> dbus-monitor --session "type='method_call',interface='org.kde.StatusNotifierWatcher',member='RegisterStatusNotifierItem'"
> ```
> 
> and share its output.

```fish
❯ dbus-monitor --session "type='method_call',interface='org.kde.StatusNotifierWatcher',member='RegisterStatusNotifierItem'"
signal time=1788655549.270079 sender=org.freedesktop.DBus -> destination=:1.436 serial=4294967295 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameAcquired
   string ":1.436"
signal time=1788655549.270094 sender=org.freedesktop.DBus -> destination=:1.436 serial=4294967295 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameLost
   string ":1.436"
method call time=1788655553.350175 sender=:1.441 -> destination=org.kde.StatusNotifierWatcher serial=15 path=/StatusNotifierWatcher; interface=org.kde.StatusNotifierWatcher; member=RegisterStatusNotifierItem
   string "org.freedesktop.StatusNotifierItem-2708788-1"
^C
[took 19.4s] ✗ 130 SIGINT · dbus-monitor --session "type='method_call',interface='org.kde.StatusNotifierWatcher',member='Registe
```

kotontrion:
According to this output that the app calls the register function with the `org.freedesktop.StatusNotifierItem-2708788-1` parameter, which does not contain a /, so your commit changes nothing in the way its handled.

If you implement a workaround for misbehaving apps not following the protocol, you need provide proof for that.

kotontrion:
Had a closer look at this.
I couldn't reproduce this, because this is patched in the Arch Linux electron package. After a quick research I was able to reproduce this. While this is already fixed in electrons next release, this is already fixed in chromium as well.
But as long as the clients adhere to the dbus naming specification, they will never end up in this added clause, so this PR does not change the behavior for valid clients. We do not provide support for client, which do not adhere to the spec, but this change is safe to merge anyway.

PR #482, opened 2026-09-07
chore(idle-notify): fix typos in docs and comments

ARKye03:
Some typos from recent PR. And the name of the protocol is …-notify. https://wayland.app/protocols/ext-idle-notify-v1

However, I understand that maybe -notifier, is the way to, I'll change it if needed.


kotontrion:
The manager object is called notifier, i guess thats where it came from.
I didn't expect it to get merged that quickly.
Thanks.

